PR #29, opened 2021-04-01
feat: add image storage to google buckets

### LINKED ISSUE #8: Investigate CDN usage
Tile image_urls should not be returned to clients, they should instead receive a url pointing to our CDN to avoid privacy concerns.

A few tasks here to investigate how this would work:

- How we upload data to a CDN (probably POSTing to some GCP storage bucket), we may not have an available Rust GCP client that handles this
- image_urls would likely not change often, but how to handle it when they do. they could potentially change/expire as often as their parent Tiles. Might need to implement something like a) Get tiles, for each tile b) check if image_url is already on CDN (and matches the version linked), adding or updating it if it does not exist or changed. Cleanup of expired images?

jrconlin:
## Description

This PR is a demo for how to add image content to a google bucket for later retrieval and use by google the cache manager. 

## Testing

This is a demo work and may not land.

## Issue(s)

Closes #8.


src/server/img_storage.rs
```diff
@@ -0,0 +1,230 @@
...
+            ..Default::default()
+        })
+        .await
+        {
+            Ok(v) => v,
+            Err(cloud_storage::Error::Google(ger)) => {
+                if ger.errors_has_reason(&cloud_storage::Reason::Conflict) {
+                    dbg!("Already exists", &sset.bucket_name);
+                    // try fetching the existing bucket.
+                    match Bucket::read(&sset.bucket_name).await {
```
pjenvey, line 64: nit: map_err would tighten this up a bit

```suggestion
                    Bucket::read(&sset.bucket_name).await.map_err(|e| {
                        HandlerError::internal(&format!("Could not read bucket {:?}", e))
                    })?;
```

src/server/img_storage.rs
```diff
@@ -0,0 +1,230 @@
...
+        // (e.g. set the path to be the SHA1 of the bytes or whatever.)
+        dbg!("fetching...", &uri);
+        /*
+        // Should we preserve the name of the image?
+        let mut hasher = Sha256::new();
+        hasher.update(uri.path().as_bytes());
+        let hash = hex::encode(hasher.finalize().as_slice());
+        */
+
+        let res = reqwest::get(&uri.to_string())
```
pjenvey, line 161: If/when we begin using this we'd want to reuse the same reqwest client (don't need it now)

src/server/img_storage.rs
```diff
@@ -0,0 +1,230 @@
...
+                        Err(e) => {
+                            return Err(HandlerErrorKind::Internal(format!(
+                                "Could not read bucket {:?}",
+                                e
+                            ))
+                            .into())
+                        }
+                    }
+                } else {
+                    return Err(HandlerErrorKind::Internal(format!(
```
pjenvey, line 80: nit: Switch from `HandlerErrorKind::Internal` to `HandlerError::internal` to save a lil boilerplate (all the `into()`'s can go away)

Cargo.toml
```diff
@@ -16,19 +16,23 @@ actix-http = "2"
...
+cloud-storage = "0.6" # 0.7+ includes request 0.11, tokio 1.4
 docopt = "1.1"
-config = "0.10"
+config = "0.11"
 env_logger = "0.8"
 futures = "0.3"
 hostname = "0.3"
 lazy_static = "1.4"
-log = { version = "0.4", features = ["max_level_debug", "release_max_level_info"] }
+log = { version = "0.4", features = ["max_level_trace", "release_max_level_info"] }
```
pjenvey, line 28: I believe logging can easily get too chatty w/ this, we probably want to revert it

src/server/mod.rs
```diff
@@ -48,6 +50,8 @@ macro_rules! build_app {
             .wrap(Cors::permissive())
             // Next, the API we are implementing
             .service(web::resource("/v1/tiles").route(web::get().to(handlers::get_tiles)))
+            // image cache tester...
```
pjenvey, line 53: Do we want this on or commented out? Or is there a `cfg(debug)` we can wrap it in?

pjenvey (changes requested):

src/server/img_storage.rs
jrconlin: Yeah, I kinda view this more as example code than anything else. There are a few good optimizations I'm not doing for the sake of legibility

Cargo.toml
jrconlin: Yeah, good catch. That was mostly to work out a problem I was hitting with reqwest.

src/server/mod.rs
jrconlin: I'll comment it out for now. Again, mostly demo code.

src/server/img_storage.rs
jrconlin: a bit, but we are potentially building out `Self` here, so having the match doesn't really impact that a lot.

pjenvey (approved):

Issue #30, opened 2021-04-08
3.1.2 Lookup country/subdivision by IP address


pjenvey:
Should be done via the MaxMind commercial db

pjenvey:
Reopening as this isn't completely done yet: 

- [x] #49 added maxmind support
- [ ] #51 plugs it in with a fallback to the load balancer headers)
- [ ] the `?country=<country>` url parameter the client specified for phase 0 should be removed

jrconlin:
Closed as OBE.

All issues addressed.

PR #49, opened 2021-04-15
Look up location via MaxMind DB

jrconlin:
## Description

This uses [the maxmind GeoLite2](https://dev.maxmind.com/geoip/geoip2/geolite2/) free database to provide IP Based location lookup information. This provides the basic interface, and I expect that further refinement will be needed.

Testing currently presumes that the database is located in `./mmdb/` directory. For operation, you can specify the path to the db file via the `maxminddb_loc` setting. 

## Testing

Unit testing included. If no GeoLite2 database is found under `./mmdb/`, location testing is skipped with a print message. 

## Issue(s)

Issue #30.

### DIFF
--- a/src/server/location.rs
+++ b/src/server/location.rs
@@ -1,0 +1,331 @@
+use std::collections::BTreeMap;
+use std::net::IpAddr;
+use std::sync::Arc;
+
+use maxminddb::{self, geoip2::City, MaxMindDBError};
+use serde::{self, Serialize};
+
+use crate::error::{HandlerErrorKind, HandlerResult};
+use crate::settings::Settings;
+
+#[derive(Serialize, Debug, Default, Clone)]
+pub struct LocationResult {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub city: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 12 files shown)


src/server/location.rs
```diff
@@ -0,0 +1,356 @@
...
+                iso_code: Some(".."),
+                names: Some({"lang": "name", ...})
+                }]),
+            traits: None }
+        }
+        */
+        let mut result = LocationResult::default();
+        match self.iploc.clone().unwrap().lookup::<City<'_>>(ip_addr) {
+            Ok(location) => {
+                if let Some(names) = location.city.and_then(|c| c.names) {
```
pjenvey, line 222: Note: I don't see us needing the language handling methods, we'd pass the country/most specific subdivision `iso_code` instead of one of `names` (though we may need to look into City's `names` if we begin using it for large enough cities).

pjenvey (approved):

src/server/location.rs
jrconlin: Yeah, the language bit is more a relic I brought over from the connection server code, mostly since MaxMind returns a set of names sorted in alphabetical by location. We *could* absolutely just force to `en` here.

pjenvey (approved):

Issue #36, opened 2021-04-09
3.2.2.1 Compose and send the adM request for tiles


jrconlin:
adM requires a GET using the following URL specification.

`https://[partner host]/tilesp` with the following query parameters:

derived values may come from the returned value of #32

| param | value | 
|--|--|
| partner |  partner name (from config) |
|sub1 | sub1 (from config) |
|sub2|  ad placement location  (enum from config) |
|country-code | IP derived country code|
|region-code | IP derived region code| 
|~~dma-code~~ |~~IP derived region code~~|
| form-factor | derived from UA | 
| os-family | derived from UA |
| v | version (from config) |

request should handle all standard error responses and prepare successful returns for further processing.

jrconlin:
Closed as part of #51

PR #51, opened 2021-04-21
feat: Add initial filtering

### LINKED ISSUE #18: 3.2.2.3 Support filtering and metadata configuration of Tiles
We need the ability to:

- restrict which tiles are presented to the client
- addition of metadata to each tile

jrconlin:
## Description

This allows strict host filtering based on a config setting `allowed_vendors`
which is a JSON array of allowed vendor host names (e.g.
`["example.com", "foo.example.org", "bar.foo.example.org"]`)
Note: currently hosts must match exactly. so for this example
"https://example.com/link" would match, but
"https://gorp.example.com/link" would not.

## Testing

TODO: Add testing

## Issue(s)

Closes #18

### DIFF
--- a/src/adm.rs
+++ b/src/adm.rs
@@ -3,8 +3,11 @@ use std::collections::{BTreeMap, HashMap};
 use url::Url;
 
 use crate::error::{HandlerError, HandlerErrorKind, HandlerResult};
+use crate::server::location::LocationResult;
 use crate::server::ServerState;
 use crate::settings::Settings;
+use crate::tags::Tags;
+use crate::web::middleware::sentry as l_sentry;
 //use crate::server::img_storage;
 
 pub(crate) const DEFAULT: &str = "DEFAULT";
@@ -38,6 +41,11 @@ pub struct AdmAdvertiserFilterSettings {
 pub(crate) type AdmSettings = HashMap<String, AdmAdvertiserFilterSettings>;
 
 impl AdmFilter {
+    /// Report the error directly to sentry
+    fn report(&self, error: &HandlerError, tags: &Tags) {
+        l_sentry::report(tags, sentry::event_from_error(error));
+    }
+
     /// Check the impression URL to see if it's valid.
     ///
     /// This extends `filter_and_process`
@@ -77,17 +85,18 @@ impl AdmFilter {
     ///
     /// - Returns None for tiles that shouldn't be shown to the client
     /// - Modifies tiles for output to the client (adding additional fields, etc.)
-    pub fn filter_and_process(&self, tile: AdmTile) -> Option<AdmTile> {
+    pub fn filter_and_process(&self, tile: AdmTile, tags: &Tags) -> Option<AdmTile> {
         let parsed: Url = match tile.advertiser_url.parse() {
             Ok(v) => v,
             Err(e) => {
-                error!(
-                    "{:?}",
-                    HandlerErrorKind::UnexpectedSiteHost(format!(
+                self.report(
+                    &HandlerErrorKind::UnexpectedSiteHost(format!(
                         "Invalid host: {:?} {:?}",
                         e,
                         tile.advertiser_url.to_string()
                     ))
+                    .into(),
+                    tags,
                 );
                 return None;
             }
@@ -95,12 +104,13 @@ impl AdmFilter {
         let host = match parsed.host() {
             Some(v) => v.to_string(),
             None => {
-                error!(
-                    "{:?}",
-                    HandlerErrorKind::Validation(format!(
+                self.report(
+                    &HandlerErrorKind::Validation(format!(
                         "Missing host from advertiser URL: {:?}",
                         parsed
                     ))
+                    .into(),
+                    tags,
                 );
                 return None;
             }
@@ -121,7 +131,7 @@ impl AdmFilter {
                 match self.check_impression(impression_filter, &mut result) {
                     Ok(_) => {}
                     Err(e) => {
-                        error!("{:?}", e);
+                        self.report(&e, tags);
                         return None;
                     }
                 }
@@ -133,7 +143,7 @@ impl AdmFilter {
                 Some(result)
             }
             None => {
-                error!("{:?}", HandlerErrorKind::UnexpectedSiteHost(host));
+                self.report(&HandlerErrorKind::UnexpectedSiteHost(host).into(), tags);
                 None
             }
         }
@@ -198,10 +208,11 @@ pub struct AdmTile {
 pub async fn get_tiles(
     reqwest_client: &reqwest::Client,
     adm_endpoint_url: &str,
-    location: (String, String), // TODO: in lieu of Location
+    location: &LocationResult,
     stripped_ua: &str,
     placement: &str,
     state: &ServerState,
+    tags: &Tags,
 ) -> Result<AdmTileResponse, HandlerError> {
     // XXX: Assumes adm_endpoint_url includes
     // ?partner=<mozilla_partner_name>&sub1=<mozilla_tag_id> (probably should
@@ -214,8 +225,8 @@ pub async fn get_tiles(
             ("partner", settings.partner_id.as_str()),
             ("sub1", settings.sub1.as_str()),
             // ("sub2", placement),
-            ("country-code", location.0.as_str()),
-            ("region-code", location.1.as_str()),
+            ("country-code", &location.country()),
+            ("region-code", &location.region()),
             // ("dma-code", location.dma),
             // ("form-factor", form_factor),
             ("os-family", stripped_ua),
@@ -244,7 +255,7 @@ pub async fn get_tiles(
     response.tiles = response
         .tiles
         .into_iter()
-        .filter_map(|tile| state.filter.filter_and_process(tile))
+        .filter_map(|tile| state.filter.filter_and_process(tile, tags))
         .collect();
     Ok(response)
 }
--- a/src/adm.rs
+++ b/src/adm.rs
@@ -146,10 +146,9 @@ impl AdmFilter {
     ///
     /// - Returns None for tiles that shouldn't be shown to the client
     /// - Modifies tiles for output to the client (adding additional fields, etc.)
-    pub fn filter_and_process(&self, tile: AdmTile, tags: &mut Tags) -> Option<AdmTile> {
+    pub fn filter_and_process(&self, mut tile: AdmTile, tags: &mut Tags) -> Option<AdmTile> {
         // Use strict matching for now, eventually, we may want to use backwards expanding domain
         // searches, (.e.g "xyz.example.com" would match "example.com")
-        let mut result = tile.clone();
         match self.filter_set.get(&tile.name.to_lowercase()) {
             Some(filter) => {
                 // Apply any additional tile filtering here.
@@ -175,23 +174,23 @@ impl AdmFilter {
                 } else {
                     filter
                 };
-                if let Err(e) = self.check_advertiser(adv_filter, &mut result, tags) {
+                if let Err(e) = self.check_advertiser(adv_filter, &mut tile, tags) {
                     self.report(&e, tags);
                     return None;
                 }
-                if let Err(e) = self.check_click(click_filter, &mut result, tags) {
+                if let Err(e) = self.check_click(click_filter, &mut tile, tags) {
                     self.report(&e, tags);
                     return None;
                 }
-                if let Err(e) = self.check_impression(impression_filter, &mut result, tags) {
+                if let Err(e) = self.check_impression(impression_filter, &mut tile, tags) {
                     self.report(&e, tags);
                     return None;
                 }
                 // Use the default.position (Option<u8>) if the filter.position (Option<u8>) isn't
                 // defined. In either case `None` is a valid return, but we should favor `filter` over
                 // `default`.
-                result.position = filter.position.or(default.position);
-                Some(result)
+                tile.position = filter.position.or(default.position);
+                Some(tile)
             }
             None => {
                 self.report(
--- a/src/adm.rs
+++ b/src/adm.rs
@@ -1,8 +1,8 @@
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, HashMap};
-use url::{Host, Url};
+use url::Url;
 
-use crate::error::{HandlerError, HandlerResult};
+use crate::error::{HandlerError, HandlerErrorKind, HandlerResult};
 use crate::server::ServerState;
 use crate::settings::Settings;
 //use crate::server::img_storage;
@@ -22,7 +22,8 @@ pub struct AdmFilter {
 #[derive(Clone, Debug, Deserialize, Default)]
 pub struct AdmAdvertiserFilterSettings {
     // valid
-    pub(crate) advertiser_url: Vec<String>,
+    pub(crate) advertiser_urls: Vec<String>,
+    pub(crate) impression_urls: Vec<String>,
     pub(crate) position: Option<u8>,
     pub(crate) include_regions: Vec<String>,
 }
@@ -30,6 +31,41 @@ pub struct AdmAdvertiserFilterSettings {
 pub(crate) type AdmSettings = HashMap<String, AdmAdvertiserFilterSettings>;
 
 impl AdmFilter {
+    /// Check the impression URL to see if it's valid.
+    ///
+    /// This extends `filter_and_process`
+    fn check_impression(
+        &self,
+        filter: &AdmAdvertiserFilterSettings,
+        tile: &mut AdmTile,
+    ) -> HandlerResult<()> {
+        let parsed: Url = match tile.impression_url.parse() {
+            Ok(v) => v,
+            Err(e) => {
+                return Err(HandlerErrorKind::UnexpectedImpressionHost(format!(
+                    "Invalid host: {:?} {:?}",
+                    e,
+                    tile.impression_url.to_string()
+                ))
+                .into());
+            }
+        };
+        let host = match parsed.host() {
+            Some(v) => v.to_string(),
+            None => {
+                return Err(HandlerErrorKind::UnexpectedImpressionHost(format!(
+                    "Missing impression host: {:?}",
+                    tile.impression_url
+                ))
+                .into());
+            }
+        };
+        if !filter.impression_urls.contains(&host) {
+            return Err(HandlerErrorKind::UnexpectedImpressionHost(host).into());
+        }
+        Ok(())
+    }
+
     /// Filter and process tiles from ADM:
     ///
     /// - Returns None for tiles that shouldn't be shown to the client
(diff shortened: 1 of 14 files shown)


src/settings.rs
```diff
@@ -1,6 +1,6 @@
 //! Application settings objects and initialization
 
-use std::collections::HashMap;
+use std::collections::BTreeMap;
```
mythmon, line 3: Why the switch from HashMap to BTreeMap (and likewise, the use of BTreeSet)?

jrconlin: For the sorts of lookups we're going to be doing, (mostly alpha based, and not a lot of entries) a BTreeMap might be a bit faster than hashing the entire string to do basically the same thing. (Tempted to also do case normalization here too, to be honest.)

pjenvey (changes requested):
I think this should filter on the top level advertiser name, e.g. instead of `CONTILE_ALLOWED_VENDORS` of `["www.acme.biz", "www.example.ninja"]` it would be 
```
{
    "Acme": {
        "advertiser_url": "www.acme.biz"
    },
    "Example": {
        "advertiser_url": "www.example.ninja"
    }
}
```

The `allowed_hosts` accomplishes this in another way but I think the host restrictions really need to be per advertiser. I think specifying advertiser name "Acme" makes it clearer what we're actually allowing/showing to clients -- I'm imagining an ad ops person managing these settings with reviewers signing off. This format somewhat resembles the current top-sites bucket we're already using.

We'll also need additional per advertiser metadata in these dicts, e.g. position and likely a list of "include_regions".

src/adm.rs
```diff
@@ -1,13 +1,79 @@
...
+        let mut allowed_url_hosts: BTreeSet<String> = BTreeSet::new();
+        if let Some(hosts) = settings.clone().allowed_vendors {
+            for host in hosts {
+                allowed_url_hosts.insert(host);
+            }
+        };
+
+        AdmFilter {
+            allowed_url_hosts: if allowed_url_hosts.is_empty() {
+                None
```
pjenvey, line 68: I'd rather this not be an `Option` but it could be useful to have an `allow all` for local debugging (that acts like the current `None` behavior here)

jrconlin: I kept debating about the `Option` as well, but it struck me that it could be possible for ADM filter to return a valid response, that contained no valid entries. If that's not really possible, then I'm fine with this returning a `Result<AdmFilter>`

pjenvey:
> Thinking a bit more about [#51 (review)](https://github.com/mozilla-services/contile/pull/51#pullrequestreview-646160332):
> Does it make sense to have the `advertiser_url` be a list, to allow for multiple potential sites? I'm also going to guess we want `position` to be a list.

Yep, I meant `advertiser_url` to be a list there. `position` shouldn't be, it's going to be added to each advertiser's tile. e.g.

```
{
    "tiles": [
        {
            "advertiser_url": "...",
            "click_url": "...",
            "id": 74161,
            "name": "Acme",
            "position": 2,
            ...
        }
    ]
}
```

jrconlin:
Right, but for the verification filter, we probably want to specify position as a list of valid locations (say if "Example" is only allowed in positions 2 and 3, but asks for position 1 or something)

pjenvey:
> Right, but for the verification filter, we probably want to specify position as a list of valid locations (say if "Example" is only allowed in positions 2 and 3, but asks for position 1 or something)

Resolved this offline -- `position` isn't for filtering, we supply a value for it for each advertiser that's simply added to Tiles outgoing to the client.

First and foremost this is an inclusion filter: e.g. include Amazon, eBay, but since we have these other per advertiser pieces of metadata, it makes sense to put also them along side the advertiser here inside the inclusion filter.

pjenvey:
^ oops

src/adm.rs
```diff
@@ -1,39 +1,225 @@
...
+            return Err(HandlerErrorKind::UnexpectedImpressionHost(host).into());
+        }
+        Ok(())
+    }
+
+    /// Filter and process tiles from ADM:
+    ///
+    /// - Returns None for tiles that shouldn't be shown to the client
+    /// - Modifies tiles for output to the client (adding additional fields, etc.)
+    pub fn filter_and_process(&self, tile: AdmTile) -> Option<AdmTile> {
```
ncloudioj, line 80: @jrconlin - Those initial filters look good 👍

We also have a concept of ["approved advertisers"](https://mana.mozilla.org/wiki/pages/viewpage.action?spaceKey=FPS&title=Mozilla+Approved+Advertisers). I think we can also check that here.

Feel free to do this in a follow-up.

src/adm.rs
```diff
@@ -1,39 +1,225 @@
...
+    pub(crate) include_regions: Vec<String>,
+}
+
+pub(crate) type AdmSettings = HashMap<String, AdmAdvertiserFilterSettings>;
+
+impl AdmFilter {
+    /// Check the impression URL to see if it's valid.
+    ///
+    /// This extends `filter_and_process`
+    fn check_impression(
```
ncloudioj, line 44: Likewise, the `click_url` can be checked as well.

src/adm.rs
jrconlin: I was thinking that list would go into the "DEFAULT" section in the configuration. I've found it's best to avoid hard coding values whenever possible.

src/adm.rs
jrconlin: Cool, refactored things a bit to reduce possible code duplication.

src/settings.rs
jrconlin: Yeah, the difference should be super minimal, and we could easily swap between the two.

src/adm.rs
```diff
@@ -1,39 +1,269 @@
...
 use url::Url;
 
-use crate::error::HandlerError;
+use crate::error::{HandlerError, HandlerErrorKind, HandlerResult};
+use crate::server::location::LocationResult;
+use crate::server::ServerState;
+use crate::settings::Settings;
+use crate::tags::Tags;
+use crate::web::middleware::sentry as l_sentry;
+//use crate::server::img_storage;
```
mythmon, line 11: Nit: should this commented out use statement be here?

src/adm.rs
```diff
@@ -1,39 +1,269 @@
...
+    /// Check the advertiser URL
+    fn check_advertiser(
+        &self,
+        filter: &AdmAdvertiserFilterSettings,
+        tile: &mut AdmTile,
+        tags: &mut Tags,
+    ) -> HandlerResult<()> {
+        check_url!(
+            &tile.advertiser_url,
+            "Advertizer",
```
mythmon, line 89: Is this spelling with a `z` intentional?

src/adm.rs
```diff
@@ -1,39 +1,269 @@
...
+            }
+        };
+        if !$filter.contains(&host) {
+            $tags.add_tag("type", $species);
+            $tags.add_extra("url", $url);
+            return Err(HandlerErrorKind::UnexpectedHost($species, host).into());
+        }
+        Ok(())
+    }};
+}
```
mythmon, line 70: Apologies if I'm missing something obvious, but why is this a macro instead of a function like this?

```rust
fn check_url(url: &str, species: &str, filter: &AdmAdvertiserFilterSettings, tags: &Tags) -> HandlerResult<()>
```

src/adm.rs
```diff
@@ -1,39 +1,269 @@
...
+
+impl AdmFilter {
+    /// Report the error directly to sentry
+    fn report(&self, error: &HandlerError, tags: &Tags) {
+        // dbg!(&error, &tags);
+        // TOOD: if not error.is_reportable, just add to metrics.
```
mythmon, line 76: Nit: TOOD -> TODO

src/adm.rs
```diff
@@ -1,39 +1,269 @@
...
+                    Err(e) => {
+                        self.report(&e, tags);
+                        return None;
+                    }
+                }
+                result.position = if filter.position.is_none() {
+                    default.position
+                } else {
+                    filter.position
+                };
```
mythmon, line 178: Could this use `unwrap_or`?

```rust
result.position = filter.position.unwrap_or(default.position);
```

mythmon (commented):
I don't really follow everything that's happening here, since I haven't been tracking this issue much. I commented some tiny nits that might be improved, and I had some questions. I don't see any problems that should prevent this from merging though.

src/adm.rs
jrconlin: ah, that's probably a hold over from the image caching PR. We're still debating if it's needed or if we can push straight to the CDN.

src/adm.rs
jrconlin: No, it's just that I constantly screw that up. Thanks!

src/adm.rs
jrconlin: because when I started i was doing more things that couldn't be done with just a function call (e.g. inlining the `$species` into the returned error names), and was change blind to the fact that it didn't need to be a macro anymore.

I'll convert it to a full function, thanks!

src/adm.rs
```diff
@@ -1,39 +1,269 @@
...
     stripped_ua: &str,
     placement: &str,
+    state: &ServerState,
+    tags: &mut Tags,
 ) -> Result<AdmTileResponse, HandlerError> {
     // XXX: Assumes adm_endpoint_url includes
     // ?partner=<mozilla_partner_name>&sub1=<mozilla_tag_id> (probably should
     // validate this on startup)
+    // let user_loc =
+    let settings = state.settings.clone();
```
ncloudioj, line 254: nit: is it possible to only clone the relevant parts (`partner_id` and `sub1`) here?

src/adm.rs
```diff
@@ -1,39 +1,269 @@
...
+    location: &LocationResult,
     stripped_ua: &str,
     placement: &str,
+    state: &ServerState,
+    tags: &mut Tags,
 ) -> Result<AdmTileResponse, HandlerError> {
     // XXX: Assumes adm_endpoint_url includes
     // ?partner=<mozilla_partner_name>&sub1=<mozilla_tag_id> (probably should
     // validate this on startup)
+    // let user_loc =
```
ncloudioj, line 253: Is this comment still needed?

src/adm.rs
jrconlin: Actually, gonna loop back on this one, because It's a bit subtle and I wrote it at 10:00pm, which is never a good thing.

I can't really use "unwrap" here because I don't actually want the unwrapped content. I want the fact that this is an `Option<u8>`. Doing the check the way I specified just lets me check if I should use the `default` value if `filter` isn't set. Both could be `None`.

src/adm.rs
```diff
@@ -54,24 +284,11 @@ pub async fn get_tiles(
         .error_for_status()?
         .json()
         .await?;
+    // TODO: Is there a better way to raise the error possibly inside of `filter_and_process`?
     response.tiles = response
         .tiles
         .into_iter()
-        .filter_map(filter_and_process)
+        .filter_map(|tile| state.filter.filter_and_process(tile, tags))
```
ncloudioj, line 291: We might want to add a `.take(settings.MAX_TILES)` here to cap the return tile set. Firefox is anticipating no more than 2 sponsored tiles for now.

src/adm.rs
jrconlin: Hrm, wondering if we should only query adm for 2 as well, or do we expect to have to toss out tiles a lot?

src/adm.rs
jrconlin: Fixed another one as well. 
(Man, my typing has gone to crap lately.)

src/adm.rs
pjenvey: You want `result.position = filter.position.or(default.position)`

src/adm.rs
```diff
@@ -1,43 +1,279 @@
...
+/// The AdmAdvertiserFilterSettings contain the settings for the various
+/// ADM provided partners. These are specified as a JSON formatted hash
+/// that contains the components. A special "DEFAULT" setting provides
+/// information that may be used as a DEFAULT, or commonly appearing set
+/// of data.
+/// See `impl From<Settings>` for details of the structure.
+#[derive(Clone, Debug, Deserialize, Default)]
+pub struct AdmAdvertiserFilterSettings {
+    pub(crate) advertiser_hosts: Vec<String>,
+    pub(crate) impression_hosts: Vec<String>,
```
pjenvey, line 36: Clicks & Impressions (`impression/click_hosts`) are _always_ going back to adM. Having a setting for a list of their valid hosts is a good idea but it shouldn't be per advertiser, it'd be global.

Whereas landing pages (`advertiser_url`/`advertiser_hosts`) take you to the advertiser itself, they need to be per advertiser as you already have it.

jrconlin: I'm fine hard-coding the "filter" to take from `default` for those if you want to skip the partner level overrides.

src/adm.rs
pjenvey: I think we'll need to query adm for a result count larger than our `max_size` due to our filtering. They're not serving very many tiles so your guess is as good as mine on how many to ask for -- a few, maybe 10.

jrconlin: settingsman

ncloudioj (dismissed):
r=nanj

This already does a LOT filtering as an "initial" patch ;). Feel free to move other comments/nits to follow-ups.

Thanks! @jrconlin

jrconlin:
@pjenvey Does [the change](https://github.com/mozilla-services/contile/pull/51/commits/16827971b3e98d7a1b246f242ec8cd4cff74d636#diff-518ba8f9a3be6880991e18c68204f793c54f128f061743d707db1c547d79b5b8R130) satisfy your request?

src/settings.rs
```diff
@@ -25,23 +26,46 @@ static DEFAULT_ADM_COUNTRY_IP_MAP: &str = r#"
...
+    /// adm country to default IP map (Hash in JSON format)
     pub adm_country_ip_map: String,
-    /// Expire tiles after this many seconds
+    /// max tiles to accept from ADM (default: 2)
+    pub adm_max_tiles: u8,
+    /// number of tiles to query from ADM (default: 10)
+    pub adm_query_tile_count: u8,
+    /// Expire tiles after this many seconds (15 * 60s)
     pub tiles_ttl: u32,
+    /// list of allowed vendors (Array in JSON format)
```
pjenvey, line 59: no longer an array

src/settings.rs
```diff
@@ -57,9 +81,14 @@ impl Default for Settings {
             actix_keep_alive: None,
             adm_endpoint_url: "".to_owned(),
             adm_country_ip_map: DEFAULT_ADM_COUNTRY_IP_MAP.to_owned(),
+            adm_max_tiles: 2,
+            adm_query_tile_count: 10,
             tiles_ttl: 15 * 60,
+            adm_settings: AdmSettings::default(),
```
pjenvey, line 87: `config` won't parse this json for us, we have to do it ourselves (see how `adm_country_ip_map` is handled).

src/web/handlers.rs
```diff
@@ -65,23 +28,45 @@ pub async fn get_tiles(
...
     let audience_key = cache::AudienceKey {
-        country: treq.country,
-        fake_ip: fake_ip.clone(),
+        country: location.country(),
+        region: location.region(),
+        // fake_ip: fake_ip.clone(),
         platform: stripped_ua.clone(),
         placement: treq.placement.clone(),
     };
+    let location = LocationResult::from(request.head());
```
pjenvey, line 69: kill this line: we already determined location

src/adm.rs
```diff
@@ -1,43 +1,275 @@
...
+    let settings = &state.settings;
     let adm_url = Url::parse_with_params(
         adm_endpoint_url,
         &[
-            ("ip", fake_ip),
-            ("ua", &stripped_ua),
-            ("sub2", &placement),
+            ("partner", settings.partner_id.as_str()),
+            ("sub1", settings.sub1.as_str()),
+            // ("sub2", placement),
```
pjenvey, line 262: kill

src/adm.rs
```diff
@@ -1,43 +1,275 @@
...
+            ("country-code", &location.country()),
```
pjenvey, line 263: Looks like you've taken care of #36 here too -- note that this will break the dev env until adM rolls out support for their new Tile API which is supposed to be mid may.

I'm checking if the dev is being used by the client team

src/adm.rs
```diff
@@ -1,43 +1,275 @@
...
+
+/// Construct the AdmFilter from the provided settings.
+/// This uses a JSON construct of settings, e.g.
+/// ```javascript
+/// /* for the Example Co advertiser... */
+/// {"Example": {
+///     /* The allowed hosts for URLs */
+///     "advertiser_hosts": ["www.example.org", "example.org"],
+///     /* Valid tile positions for this advertiser (empty for "all") */
+///     "positions": [1, 2],
```
pjenvey, line 203: this should be singular `position`

src/adm.rs
```diff
@@ -1,43 +1,275 @@
...
+        let mut result = tile.clone();
+        match self.filter_set.get(&tile.name.to_lowercase()) {
+            Some(filter) => {
+                // Apply any additional tile filtering here.
+                let none = AdmAdvertiserFilterSettings::default();
+                let default = self
+                    .filter_set
+                    .get(&DEFAULT.to_lowercase())
+                    .unwrap_or(&none);
+                let adv_filter = if filter.advertiser_hosts.is_empty() {
```
pjenvey, line 142: looks like these could all use the `Option::or` like `result.position` below

pjenvey (changes requested):
Does this keep getting larger? Sorry, I still have some more to look at

src/web/handlers.rs
```diff
@@ -65,23 +28,45 @@ pub async fn get_tiles(
             .expect("Invalid ADM_COUNTRY_IP_MAP setting")
     };
     let stripped_ua = user_agent::strip_ua(&treq.ua);
+    let header = request.head();
+    let c_info = request.connection_info();
+    let ip_addr_str = c_info.remote_addr().unwrap_or(fake_ip);
```
pjenvey, line 33: using the `fake_ip` here is dubious but I'm not sure what we should do otherwise.

pjenvey (changes requested):

src/settings.rs
pjenvey: Pity there is no easy way of testing this (parsing of env vars completely isolated from other tests)

src/adm.rs
```diff
@@ -1,43 +1,275 @@
...
+    }
+
+    /// Filter and process tiles from ADM:
+    ///
+    /// - Returns None for tiles that shouldn't be shown to the client
+    /// - Modifies tiles for output to the client (adding additional fields, etc.)
+    pub fn filter_and_process(&self, tile: AdmTile, tags: &mut Tags) -> Option<AdmTile> {
+        // Use strict matching for now, eventually, we may want to use backwards expanding domain
+        // searches, (.e.g "xyz.example.com" would match "example.com")
+        let mut result = tile.clone();
```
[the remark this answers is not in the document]
jrconlin: I'm not so sure. `filter_map` wants `Some(value)` of of this, which would mean either cloning further into the fn or doing weird reference tricks.

src/adm.rs
jrconlin: I don't think `Vec` has an `.or()`. I could do a `filter.advertiser_hosts.append(default.clone())`, but that alters the behavior to always include the `default` instead of "use default if they haven't defined anything".

src/adm.rs
jrconlin: Just using mostly what was there.
I'll comment out `country` for now.

src/settings.rs
jrconlin: Hrm, I should also fix up the storage settings too then, thanks!

src/adm.rs
pjenvey: My bad, I read these as Options

src/adm.rs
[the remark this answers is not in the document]
pjenvey: We own `tile` so we could return it

src/web/handlers.rs
pjenvey: Let's keep it as is and just log an issue for it. If we go with the LB headers to start with we also need a similar "what should we do" when the LB returns an empty string for country/region.

src/adm.rs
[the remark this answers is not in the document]
jrconlin: Yeah, you're right. For some stupid reason I was thinking we're passing down a `&mut` instead of just `mut`.

src/adm.rs
pjenvey: Commenting out just country won't help, the integration still won't work (temporarily) unless we restore ip/ua. I haven't heard back from the client dev yet either, though.

pjenvey:
Looks good I'm just not sure which we should go w/ the adM integration piece until I hear back from the client dev

src/adm.rs
```diff
@@ -1,43 +1,287 @@
...
         adm_endpoint_url,
         &[
-            ("ip", fake_ip),
-            ("ua", &stripped_ua),
-            ("sub2", &placement),
+            ("partner", settings.partner_id.as_str()),
+            ("sub1", settings.sub1.as_str()),
+            ("ip", &location.fake_ip), // TODO: remove once ADM API finalized
+            ("ua", &stripped_ua),      // TODO: remove once ADM API finalized
+            //("country-code", &location.country()),
```
pjenvey, line 275: One last thing: let's put country back (for the new API)

pjenvey (changes requested):
^

pjenvey (approved):

Issue #48, opened 2021-04-13
3.2.2.5 Cache UA response data


jrconlin:
In order to ensure high performance, improve privacy, and reduce demand on the adM servers, we should cache responses where possible. This will involve determining how to optimally structure the cache, the data to cache and the period we should cache.

jrconlin:
closed as part of #76

PR #76, opened 2021-05-06
Feat/48 cache2

jrconlin:
## Description

Use a Hash of selected values in the AdmTile rather than the JSONized string.

## Issue(s)

Issue #48

### DIFF
--- a/src/adm.rs
+++ b/src/adm.rs
@@ -1,17 +1,18 @@
 use serde::{Deserialize, Serialize};
-use std::hash::{Hash, Hasher};
 use std::{
     collections::{hash_map::DefaultHasher, HashMap},
     fmt::Debug,
+    hash::{Hash, Hasher},
 };
 use url::Url;
 
-use crate::error::{HandlerError, HandlerErrorKind, HandlerResult};
-use crate::server::location::LocationResult;
-use crate::server::ServerState;
-use crate::settings::Settings;
-use crate::tags::Tags;
-use crate::web::middleware::sentry as l_sentry;
+use crate::{
+    error::{HandlerError, HandlerErrorKind, HandlerResult},
+    server::{location::LocationResult, ServerState},
+    settings::Settings,
+    tags::Tags,
+    web::middleware::sentry as l_sentry,
+};
 //use crate::server::img_storage;
 
 pub(crate) const DEFAULT: &str = "DEFAULT";
--- a/src/adm.rs
+++ b/src/adm.rs
@@ -1,5 +1,9 @@
 use serde::{Deserialize, Serialize};
-use std::{collections::HashMap, fmt::Debug};
+use std::hash::{Hash, Hasher};
+use std::{
+    collections::{hash_map::DefaultHasher, HashMap},
+    fmt::Debug,
+};
 use url::Url;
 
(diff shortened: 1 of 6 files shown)


src/server/cache.rs
```diff
@@ -86,25 +88,35 @@ async fn tile_cache_updater(state: &ServerState) {
         match result {
             Ok(response) => {
                 //trace!("tile_cache_updater: {:#?}", response);
-                let tiles = match serde_json::to_string(&response) {
-                    Ok(tiles) => tiles,
+                let json_tiles = match serde_json::to_string(&response) {
```
pjenvey, line 91: Until we know more about how large this cache will eventually be, in retrospect, storing solely the json in it is really a premature optimization on my part.

What do you think about just storing the `AdmTileResponse` struct and having it derive `Eq/PartialEq` (`AdmTile` needs them too). We could store the rendered json along side it too (maybe that's even premature before benchmarks)  -- but I'd rather we stick to comparing the cache values with strict equality (via `AdmTileResponse`) for now vs relying on a hash code (and potential collisions).

Maybe we'll need something like this later if the cache is too large but I'm leaning towards keeping this dead simple until we need something different.

src/adm.rs
```diff
@@ -252,6 +257,18 @@ pub struct AdmTile {
     pub position: Option<u8>,
 }
 
+impl AdmTile {
```
pjenvey, line 260: FYI - this doesn't include `id` (or `position` but that's not very relevant) -- `#[derive(Hash)]` would have included them all and gave you the hash impl for free (only difference is the caller needs to create a DefaultHasher itself to pass to every `hash` method call)

src/server/location.rs
```diff
@@ -32,11 +40,13 @@ impl From<&RequestHead> for LocationResult {
         if let Some(loc_string) = headers.get(GOOG_LOC_HEADER) {
             dbg!("Found google header", loc_string);
             let mut parts: Vec<&str> = loc_string.to_str().unwrap_or("").split(',').collect();
-            return Self {
-                city: parts.pop().map(ToOwned::to_owned),
-                country: parts.pop().map(ToOwned::to_owned),
+            let reply = Self {
+                country: parts.pop().map(|v| v.trim().to_owned()),
```
pjenvey, line 44: Also note the changes here I made in #78 --

- don't want city (yet), just country and subdivision
- it's gonna be "--custom-request-header 'X-Client-Geo-Location: {client_region},{client_region_subdivision}"
- `client_region` is mostly country. `client_region_subdivision` is a "Unicode CLDR subdivision ID" which includes a 2 letter country prefix we strip

src/server/cache.rs
jrconlin: I'd be ok with using the `Eq/PartialEq` trait. I guess I was more thinking that we want to compare elements of the tile. (e.g. we may not need to compare `id` or `position` for instance.) This is kinda per the other comments. Honestly, i was more figuring that there could be variations in the returned tile values that might churn the cache and this way we'd have a normalization point.

src/server/location.rs
jrconlin: I'll correct the header we get location from.

jrconlin:
closing for reasons

Issue #60, opened 2021-05-05
Full integration testing


data-sync-user:
Create a suite of full integration tests for Contile to pull content from an adM server and provide UA ready data.

See also: https://github.com/mozilla-services/contile/issues/43

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-83)

data-sync-user:
➤ Ron Manning commented:

Ideally, this is working before adM starts testing on it. We expect adM to have their API to be production-ready by May 20

data-sync-user:
➤ JR Conlin commented:

Possible dupe of https://jira.mozilla.com/browse/CONSVC-45 ?

data-sync-user:
➤ Ron Manning commented:

Since this one is part of Phase 2, let's keep this and close CONSVC-45.

PR #79, opened 2021-05-13
Feat/60 int test

jrconlin:
## Description

Adds a simple integration testing framework, along with testing code
changes, to allow extended operational testing of the Contile server

## Testing

See `tools/test/README.md`

## Issue(s)

Issue #60

### DIFF
--- a/adm_settings_test.json
+++ b/adm_settings_test.json
@@ -1,0 +1,17 @@
+{"DEFAULT": {
+    "advertiser_hosts": ["example.com"],
+    "impression_hosts": ["example.net"],
+    "click_hosts":["example.com"]
+},
+"Acme": {
+    "advertiser_hosts": ["www.acme.biz", "acme.biz"],
+    "position": 0
+},
+"Dunder Mifflin": {
+    "advertiser_hosts": ["www.dunderm.biz", "dunderm.biz"],
+    "position": 1
+},
+"Los Pollos Hermanos": {
+    "advertiser_hosts": ["www.lph-nm.biz", "lph-mx.co.mx"]
+}
+}
--- a/src/adm/settings.rs
+++ b/src/adm/settings.rs
@@ -15,26 +15,32 @@ pub(crate) const DEFAULT: &str = "DEFAULT";
 /// These are specified as a JSON formatted hash
 /// that contains the components. A special "DEFAULT" setting provides
 /// information that may be used as a DEFAULT, or commonly appearing set
-/// of data.
+/// of data. Any Optional value that is not defined will use the value
+/// defined in DEFAULT.
 #[derive(Clone, Debug, Deserialize, Default, Serialize)]
 pub struct AdmAdvertiserFilterSettings {
-    /// Set of valid hosts for the `advertiser_url`
+    /// Required set of valid hosts for the `advertiser_url`
     pub(crate) advertiser_hosts: Vec<String>,
-    /// Set of valid hosts for the `impression_url`
+    /// Optional set of valid hosts for the `impression_url`
     #[serde(default)]
     pub(crate) impression_hosts: Vec<String>,
-    /// Set of valid hosts for the `click_url`
+    /// Optional set of valid hosts for the `click_url`
     #[serde(default)]
     pub(crate) click_hosts: Vec<String>,
-    /// valid position for the tile
+    /// Optional valid position for the tile
     pub(crate) position: Option<u8>,
-    /// Set of valid regions for the tile (e.g ["en", "en-US/TX"])
+    /// Optional set of valid regions for the tile (e.g ["en", "en-US/TX"])
     #[serde(default)]
     pub(crate) include_regions: Vec<String>,
 }
 
 pub(crate) type AdmSettings = HashMap<String, AdmAdvertiserFilterSettings>;
 
+/// Attempt to read the AdmSettings as either a path to a JSON file, or as a JSON string.
+///
+/// This allows `CONTILE_ADM_SETTINGS` to either be specified as inline JSON, or if the
+/// Settings are too large to fit into an ENV string, specified in a path to where the
+/// settings more comfortably fit.
 impl From<&Settings> for AdmSettings {
     fn from(settings: &Settings) -> Self {
         if settings.adm_settings.is_empty() {
@@ -78,7 +84,7 @@ impl From<&Settings> for HandlerResult<AdmFilter> {
     fn from(settings: &Settings) -> Self {
         let mut filter_map: HashMap<String, AdmAdvertiserFilterSettings> = HashMap::new();
         for (adv, setting) in AdmSettings::from(settings) {
-            dbg!("Processing records for {:?}", &adv.to_lowercase());
+            dbg!("Processing records for {:?}", &adv);
             // map the settings to the URL we're going to be checking
             filter_map.insert(adv.to_lowercase(), setting);
         }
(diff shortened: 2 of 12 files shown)


src/adm/settings.rs
```diff
@@ -21,12 +21,15 @@ pub struct AdmAdvertiserFilterSettings {
     /// Set of valid hosts for the `advertiser_url`
     pub(crate) advertiser_hosts: Vec<String>,
```
ethowitz, line 22: Do we need a `#[serde(default)]` tag for this attribute as well?

src/settings.rs
```diff
@@ -97,13 +101,18 @@ impl Default for Settings {
...
         }
     }
 }
 
 impl Settings {
     /// Load the settings from the config file if supplied, then the environment.
-    pub fn with_env_and_config_file(filename: &Option<String>) -> Result<Self, ConfigError> {
+    pub fn with_env_and_config_file(
+        filename: &Option<String>,
+        debug: Option<bool>,
```
ethowitz, line 114: Can this just be `bool`?

adm_settings_test.json
```diff
@@ -0,0 +1,17 @@
...
+},
+"Acme": {
+    "advertiser_hosts": ["www.acme.biz", "acme.biz"],
+    "position": 0
+},
+"Dunder Mifflin": {
+    "advertiser_hosts": ["www.dunderm.biz", "dunderm.biz"],
+    "position": 1
+},
+"Los Pollos Hermanos": {
```
ethowitz, line 14: I appreciate your choice of fictional advertisers

src/server/img_storage.rs
```diff
@@ -213,30 +213,4 @@ impl StoreImage {
             }
         }
     }
-
-    /// Download an image from a given [uri::Uri] and store it to Google Cloud Storage
-    pub async fn fetch(&self, uri: &uri::Uri) -> HandlerResult<Option<StoreResult>> {
```
ethowitz, line 218: Was the removal of this function intentional?

src/web/handlers.rs
```diff
@@ -71,14 +71,15 @@ pub async fn get_tiles(
...
         os_family,
     };
-    if let Some(tiles) = state.tiles_cache.read().await.get(&audience_key) {
-        trace!("get_tiles: cache hit: {:?}", audience_key);
-        metrics.incr("tiles_cache.hit");
-        return Ok(HttpResponse::Ok()
-            .content_type("application/json")
-            .body(&tiles.json));
+    if !state.settings.test_mode {
+        if let Some(tiles) = state.tiles_cache.read().await.get(&audience_key) {
```
ethowitz, line 75: Just making sure I understand: we don't want to hit the cache when running the integration tests? If that's true, do we still want to write to it when the server is in test mode?

tools/test/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+This directory contains a simple integration test for the Contile server.
+
+## Installation
+
+1) First, create a virtualenv installation of python
+`python -m venv venv`
+
+This will create a local install of the python. You can then "activate" it by calling
+`sh venv/bin/activate.sh`
```
ethowitz, line 11: Should this be `source venv/bin/activate`? I didn't seem to have a `venv/bin/activate.sh` file 🤔

tools/test/README.md
```diff
@@ -0,0 +1,41 @@
...
+You can run the tests by running
+```pytest . ```
+
+You can specify `pytest -sx . ` if you want tests to stop at the first failure.
+
+The test will attempt to start the locally compiled Contile server (unless `CONTILE_TEST_NOSERVER` is set) and run the local tests checking for returned values.
+
+## Crafting tests
+
+Tests are included in the `TestAdm` class. Please note that returned values from a live server may differ significantly from the test data, so examining the return results may need be blocked by checking if `settings.get("noserver")` is not set.
```
ethowitz, line 37: I think there my be a typo here ⌨️  Should the last clause read `may need **to** be blocked by checking if settings.get("noserver") is not set.`?

src/adm/settings.rs
```diff
@@ -37,7 +40,12 @@ impl From<&Settings> for AdmSettings {
         if settings.adm_settings.is_empty() {
             return Self::default();
         }
-        serde_json::from_str(&settings.adm_settings).expect("Invalid ADM Settings")
+        if Path::new(&settings.adm_settings).exists() {
+            if let Ok(f) = File::open(&settings.adm_settings) {
+                return serde_json::from_reader(f).expect("Invalid ADM Settings file");
+            }
```
ethowitz, line 46: Do you think we should return an error if the file open fails? Otherwise we'll try to read it as a JSON string even though the path does in fact exist (since we passed the check on L43). Might not be worth it, but thought I'd mention it 🤷🏻

ethowitz (changes requested):

src/adm/settings.rs
jrconlin: No. While the other values are going to be optional (and thus potentially fall back to a `Default`) we want these to be expressly defined. I can improve the comment descriptor to indicate that.

adm_settings_test.json
jrconlin: blame @pjenvey

src/web/handlers.rs
jrconlin: Having the server read from the cache means that we have to wait until the cache refreshes or is cleared before we can execute the next test. I'm ok with writing to the cache anyway because this way we're not impacting more of the code. 
I'm honestly debating if in the future we want to make reading from the cache a header designated option as well, but that could introduce subtle bugs if we run tests in parallel.

tools/test/README.md
jrconlin: either works, but yeah, that's the proper term.

src/server/img_storage.rs
jrconlin: Yes. It's equivalent to the `store()` function.

src/adm/settings.rs
jrconlin: Not sure I follow.

If `adm_settings` is empty, we use the default.

If `adm_settings` resolves to a valid file path, then we read the content of that file as the `adm_settings`.

If it's not a valid path (because it's a JSON blob), try reading that blob as JSON and use that as `adm_settings`. 

Otherwise give up with a `panic!`

jrconlin: I'll add a comment that makes that clearer. Thanks!

ethowitz: Ah I see, I was talking about the case where the file path is valid but `File::open` returns an `Err`, but I guess that's pretty unlikely since we're checking `Path::exists` first

src/adm/tiles.rs
```diff
@@ -65,21 +98,34 @@ pub async fn get_tiles(
...
-        })?
-        .error_for_status()?
-        .json()
-        .await
-        .map_err(|e| {
-            // ADM servers are not returning correct information
-            HandlerErrorKind::BadAdmResponse(format!("ADM provided invalid response: {:?}", e))
-        })?;
+    let mut response: AdmTileResponse = if state.settings.test_mode {
+        let default = HeaderValue::from_str(DEFAULT).unwrap();
```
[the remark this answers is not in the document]
jrconlin: That error pops when it can't find the test JSON file, so more than likely the `CONTILE_TEST_FILE_PATH` needed to be adjusted because you're not running in the project root. I added a better comments as well as normalized the file name to lower case to try and catch that.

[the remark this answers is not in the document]
ethowitz: Ah, I misunderstood what was going on. Thanks for clarifying!

ethowitz (dismissed):

pjenvey (approved):

PR #80, opened 2021-05-13
bug: Add additional verification checks.

### LINKED ISSUE #22: 3.3 Validate AdM Tile data
As @jrconlin [points out](https://github.com/mozilla-services/contile/pull/16#pullrequestreview-595504930): 

We might want to do some validation on a few of these fields, partly because it's external data that we're displaying on user machines, so there's a potential for mischief.

We can start w/ basic validation of the urls via the validator crate's [#[validate(url)]](https://github.com/Keats/validator#url)

jrconlin:
Closes #22

## Description

Perform additional field validation checks for `impression_url` to ensure required values are present.

## Testing

Test included

## Issue(s)

Closes #22


src/web/test.rs
```diff
@@ -56,11 +56,11 @@ macro_rules! init_app {
 }
 
 /// Bind a mock of the AdM Tiles API to a random port on localhost
-fn init_mock_adm() -> (dev::Server, SocketAddr) {
-    let server = HttpServer::new(|| {
-        App::new().route(
+fn init_mock_adm(response: String) -> (dev::Server, SocketAddr) {
+    let server = HttpServer::new(move || {
+        App::new().data(response.clone()).route(
```
jrconlin, line 61: Yes, it's gross.
Yes, it works.

src/adm.rs
```diff
@@ -123,7 +112,36 @@ impl AdmFilter {
...
         tags: &mut Tags,
     ) -> HandlerResult<()> {
-        check_url(&tile.click_url, "Click", &filter.click_hosts, tags)
+        let url = &tile.click_url;
+        let species = "Click";
+        tags.add_tag("type", species);
+        tags.add_extra("url", &url);
+        // Check the required fields are present for the `click_url`
+        // pg 15 of 5.7.21 spec.
+        let mut check_keys = vec![
```
pjenvey, line 121: nit: don't need `to_owned` on these string values due to String having `PartialEq<&str>` -- it could also be a plain array instead of vec and could even be static (lazy_static)

src/adm.rs
```diff
@@ -123,7 +112,36 @@ impl AdmFilter {
...
+            "ctag".to_owned(),
+            "version".to_owned(),
+            "key".to_owned(),
+            "ci".to_owned(),
+            "aespFlag".to_owned(),
+        ];
+        check_keys.sort();
```
pjenvey, line 128: Looks like the latest API doc has one more probably new addition here, an optional "click-status" field that we should allow if present. It's only for clicks.

pjenvey (changes requested):

pjenvey:
Also we can fix these after merging if they'll cause 3 PRs in merge hell

src/adm.rs
```diff
@@ -62,35 +62,17 @@ impl From<&Settings> for AdmSettings {
...
+            return Err(HandlerErrorKind::MissingHost(species, url.to_string()).into());
         }
     };
     if !filter.contains(&host) {
-        tags.add_tag("type", species);
-        tags.add_extra("url", &url);
         return Err(HandlerErrorKind::UnexpectedHost(species, host).into());
     }
-    Ok(())
+    Ok(true)
```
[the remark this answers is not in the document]
jrconlin: Yep. it's used as a bool in `check_click`.

src/adm.rs
pjenvey: e.g.

```rust
lazy_static! {
    static ref CLICK_PARAMS: [&'static str; 5] = {
        let mut params = [
            "ctag",
            "version",
            "key",
            "ci",
            "aespFlag",
        ];
        params.sort();
        params
    };
}
```

```
query_keys == *CLICK_PARAMS
```

src/adm.rs
jrconlin: ok. I'll add a required and optional param check.

src/adm.rs
jrconlin: I thought about that, but I wanted to keep things together as much as possible, and `lazy_static!` doesn't want to be in a `fn`.

jrconlin:
I've bought a bungalow in merge hell. 

The sunsets are nice

if you ignore the wailing and gnashing of teeth.

src/adm.rs
```diff
@@ -135,12 +175,19 @@ impl AdmFilter {
...
+        tags.add_tag("type", species);
+        tags.add_extra("url", &url);
+        let parsed: Url = match url.parse() {
+            Ok(v) => v,
+            Err(e) => {
+                tags.add_extra("parse_error", &e.to_string());
+                return Err(HandlerErrorKind::InvalidHost(species, url.to_string()).into());
+            }
+        };
+        check_url(parsed, species, &filter.impression_hosts)?;
```
pjenvey, line 189: Sorry, we forgot one more thing: we should require an `id` param on `impression_url`s

pjenvey (changes requested):

pjenvey (approved):

PR #97, opened 2021-05-21
feat: minor updates per the latest Tiles API

### LINKED ISSUE #96: adM new API integration Part 2
A couple minor updates to the Tiles API:

- `click_url`'s `aespFlag` parameter was deprecated and is no longer sent
- `os-family` is not currently used as a targeting parameter so it doesn't belong in the audience (cache) key. this could change later, but that's the case for now

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-122)

pjenvey:
## Description

- click_url's aespFlag parameter is deprecated
- note os-family isn't currently for targeting
- include more metadata in sentry errors

## Testing

Fixes dev/stage to return tiles

## Issue(s)

Closes #96

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -1,5 +1,10 @@
-use std::{collections::HashMap, fmt::Debug};
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Debug,
+    iter::FromIterator,
+};
 
+use lazy_static::lazy_static;
 use url::Url;
 
 use super::{AdmAdvertiserFilterSettings, AdmTile, DEFAULT};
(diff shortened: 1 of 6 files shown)


src/adm/filter.rs
```diff
@@ -81,48 +97,42 @@ impl AdmFilter {
...
                 return Err(HandlerErrorKind::InvalidHost(species, url.to_string()).into());
             }
         };
-        let mut query_keys = parsed
+        let query_keys = parsed
             .query_pairs()
             .map(|p| p.0.to_string())
-            .collect::<Vec<String>>();
-        query_keys.sort();
+            .collect::<HashSet<String>>();
```
pjenvey, line 115: not that sets matter much here, mostly just killing the unneeded sort() and moving it to `lazy_static`

jrconlin (approved):

PR #102, opened 2021-05-24
feat: Add a setting which defines a custom location header

### LINKED ISSUE #101: Use an optional `test header` for location lookup (for integration testing)
We currently do not have a way to test custom locations easily (either using the IP or a cloud service provided header value). We should be able to pass a custom header to specify an alternate location for testing.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-129)

jrconlin:
Closes #101

## Description

This adds a new setting: `location_test_header` which can be used to
specify a custom, testing location request header (e.g. `Test-Location`
which can contain a false test location for the request. This flag is
not meant for use in a production environment. If the header is not
present, the server will fall back on the normal location determination
code.

## Testing

unit test included.

## Issue(s)

Closes #101

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -25,6 +25,7 @@ pub async fn get_tiles(
     trace!("get_tiles");
 
     let cinfo = request.connection_info();
+    let settings = state.settings.clone();
     let ip_addr_str = cinfo.remote_addr().unwrap_or({
         let default = state
             .adm_country_ip_map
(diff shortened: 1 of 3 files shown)


src/web/handlers.rs
```diff
@@ -25,6 +25,7 @@ pub async fn get_tiles(
     trace!("get_tiles");
 
     let cinfo = request.connection_info();
+    let settings = state.settings.clone();
```
pjenvey, line 28: nit: we don't need a clone here, see [my branch](https://github.com/mozilla-services/contile/compare/feat/101-loc-head...feat/101-loc-header-ref)

jrconlin: also fixing the clippy that didn't trigger on my system.

pjenvey (approved):

Issue #92, opened 2021-05-20
Define Contile HTTP status codes


data-sync-user:
Let's use this ticket to specify the HTTP status code for Contile, then we can add the specs to the Contile technical [doc](https://docs.google.com/document/d/12hV_4b6Ur_rpdk1ic7OYOVL_v6i_F886RGXtch60-qg/edit?usp=sharing).

We should cover the following cases:

* Success - Contile was able to answer user's request with at least one tile either from cache or adM
* Failure case (1) - All the tiles from adM failed to pass the validation
* Failure case (2) - Contile or adM errored with 500
* Failure case (3) - Contile received other error codes (such as 40X) from adM

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-117)

jrconlin:
I'm going to suggest the following:

Success (with tiles): 200
Failure 1 - no tiles: 204 NO_CONTENT
Failure 2 - adm 500 error: 502 BAD_GATEWAY
Failure 3 - adm general error: 503 SERVICE_UNAVAILABLE.

The last may be a bit odd, since you normally also provide a "Retry-After:" response header. We still could, and just provide a Period + Random second value to alleviate thundering herd problems.

data-sync-user:
➤ Nan Jiang commented:

Firefox currently treats the HTTP 200 with an empty tile array as "we don't have anything to show right now", hence it won't show any sponsored tiles even if there are some from the last response.

On the other hand, if Contile responds the request with any error code, Firefox will only record an error and leave the existing sponsored tiles (if any) alone.  

I think there is some kind of ambiguity between "have nothing to show" vs "we do have something from adM, but they didn't the validation, maybe you can reuse the existing ones until we get those new ones fixed". Either way, let's define a set of status codes for Contile so that Firefox devs can reference.

jrconlin:
hrm, not to bikeshed too much, but tempted to switch failure 2 and failure 3 returns 
(so an ADM 500 would result in Contile returning a `503 SERVICE UNAVAILABLE` and a non 500 ADM error would return a `502 BAD GATEWAY`)

The rationale being that an ADM 500 would clear faster than one of the other errors, but again, this feels like bike-shedding.

data-sync-user:
➤ Nan Jiang commented:

[~jconlin@mozilla.com] - I like your proposal.

[...]

PR #104, opened 2021-05-24
feat: Add improved response codes

jrconlin:
Closes: #92

## Description

Add improved response codes:
200 - tiles included
204 - No tile content included
502 - ADM returned a 500 
503 - ADM returned a general error

## Testing

Unit test included

## Issue(s)

Issue #92

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -1,6 +1,5 @@
 //! API Handlers
 use actix_web::{web, HttpRequest, HttpResponse};
-use serde_json::json;
 
 use super::user_agent;
 use crate::{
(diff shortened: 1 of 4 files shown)


src/web/handlers.rs
```diff
@@ -92,6 +93,9 @@ pub async fn get_tiles(
     {
         Ok(response) => {
             // adM sometimes returns an invalid response. We don't want to cache that.
+            if response.tiles.is_empty() {
+                reply = HttpResponse::NoContent();
```
pjenvey, line 97: These 204's are including a payload though, of the empty set `{"tiles":[]}`, which violates the RFC.

We should return no payload in this case but let's also ensure the client side's happy with that, cc @ncloudioj

pjenvey (changes requested):

pjenvey (approved):

PR #113, opened 2021-05-26
feat: relax click/impression host checking

### LINKED ISSUE #109: Relax strict hostname checking of click/impression urls
We should allow matching click/impression urls by simply domain names.

We're only officially provided with plain domain names to match against e.g. "example.com" but the inclusion filter requires the full subdomains spelled out ("bridge.foo.bar.example.com") or the matching fails. Let's fix this now so we don't have to catch every single variant of these hostnames that could crop up over time.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-140)

pjenvey:
## Description

allow partial matching to the domains specified in the inclusion filter

e.g. "click_hosts": ["example.com"] now allows "foo.example.com"

## Testing

How should reviewers test?

## Issue(s)

Closes #109

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -262,14 +262,14 @@ mod tests {
         assert!(check_url(
             "https://example.com".parse().unwrap(),
             species,
-            &vec![vec!["example".to_owned(), "com".to_owned()]]
+            &[vec!["example".to_owned(), "com".to_owned()]]
         )
         .unwrap());
 
         assert!(check_url(
             "https://foo.bridge.example.com/?quux=baz".parse().unwrap(),
             species,
-            &vec![vec!["example".to_owned(), "com".to_owned()]]
+            &[vec!["example".to_owned(), "com".to_owned()]]
         )
         .unwrap());
     }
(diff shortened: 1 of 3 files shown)


src/adm/filter.rs
```diff
@@ -38,18 +38,28 @@ pub struct AdmFilter {
...
-fn check_url(url: Url, species: &'static str, filter: &[String]) -> HandlerResult<bool> {
-    let host = match url.host() {
-        Some(v) => v.to_string(),
-        None => {
-            return Err(HandlerErrorKind::MissingHost(species, url.to_string()).into());
+/// Extract the host from Url
+fn get_host(url: Url, species: &'static str) -> HandlerResult<String> {
+    url.host()
+        .map(|host| host.to_string())
+        .ok_or_else(|| HandlerErrorKind::MissingHost(species, url.to_string()).into())
```
mythmon, line 45: Out of curiosity, what's the etymology of the name `species` here?

src/adm/filter.rs
```diff
@@ -237,3 +251,48 @@ impl AdmFilter {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
mythmon, line 256: It's pretty clear that this bug doesn't exist now, but I think we should add a test that checks that an allowed host of "foo.com" doesn't allow "evilfoo.com". ie: it doesn't match when considering domain parts, but would match when (wrongly) considering string comparisons. It would be embarrassing for a bug like that to sneak in.

src/adm/filter.rs
pjenvey: @jrconlin's knack for naming things?

jrconlin: pretty much. Had to come up with the variant for what sort of host we're checking. "species" seemed reasonable, (as would "category", "classification", or some other version out of a thesaurus). I'm happy to accept any better suggestion.

jrconlin (approved):

src/adm/filter.rs
mythmon: I have no better suggestions, just curious.

PR #115, opened 2021-05-27
feat: use default location for unlocatable IPs

### LINKED ISSUE #95: Handle location lookup failing
Currently `LocationResult` returns empty strings for country/subdivision in the case that their lookup failed.

We don't want to emit empty strings anywhere (to adM, in our cache keys or to metrics/sentry) and the adM response w/ one would likely be an error.

There's not much we can do when lookup fails, I think we should probably:

1. default to some common location (say US-NY) so something is shown
2. emit a metric about the lookup failure

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-118)

jrconlin:
## Description

Use `CONTILE_DEFAULT_LOCATION` to specify a default location other
than `USOK`.

## Testing

includes unit test

## Issue(s)

Closes #95

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -76,7 +76,7 @@ impl AdmFilter {
         &self,
         filter: &AdmAdvertiserFilterSettings,
         tile: &mut AdmTile,
-        tags: &mut Tags,
+        mut tags: Tags,
     ) -> HandlerResult<()> {
         let url = &tile.advertiser_url;
         let species = "Advertiser";
@@ -106,7 +106,7 @@ impl AdmFilter {
         &self,
         filter: &AdmAdvertiserFilterSettings,
         tile: &mut AdmTile,
-        tags: &mut Tags,
+        mut tags: Tags,
     ) -> HandlerResult<()> {
         let url = &tile.click_url;
         let species = "Click";
@@ -160,7 +160,7 @@ impl AdmFilter {
         &self,
         filter: &AdmAdvertiserFilterSettings,
         tile: &mut AdmTile,
-        tags: &mut Tags,
+        mut tags: Tags,
     ) -> HandlerResult<()> {
         let url = &tile.impression_url;
         let species = "Impression";
@@ -221,17 +221,17 @@ impl AdmFilter {
                 } else {
                     filter
                 };
-                if let Err(e) = self.check_advertiser(adv_filter, &mut tile, tags) {
+                if let Err(e) = self.check_advertiser(adv_filter, &mut tile, tags.clone()) {
                     dbg!("bad adv");
                     self.report(&e, tags);
                     return None;
                 }
-                if let Err(e) = self.check_click(click_filter, &mut tile, tags) {
+                if let Err(e) = self.check_click(click_filter, &mut tile, tags.clone()) {
                     dbg!("bad click");
                     self.report(&e, tags);
                     return None;
                 }
-                if let Err(e) = self.check_impression(impression_filter, &mut tile, tags) {
+                if let Err(e) = self.check_impression(impression_filter, &mut tile, tags.clone()) {
                     dbg!("bad imp");
                     self.report(&e, tags);
                     return None;
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -76,16 +76,16 @@ impl AdmFilter {
         &self,
         filter: &AdmAdvertiserFilterSettings,
         tile: &mut AdmTile,
-        mut tags: Tags,
+        tags: &mut Tags,
     ) -> HandlerResult<()> {
         let url = &tile.advertiser_url;
         let species = "Advertiser";
-        tags.add_tag("type", species);
-        tags.add_extra("tile", &tile.name);
-        tags.add_extra("url", &url);
         let parsed: Url = match url.parse() {
             Ok(v) => v,
             Err(e) => {
+                tags.add_tag("type", species);
+                tags.add_extra("tile", &tile.name);
+                tags.add_extra("url", &url);
                 tags.add_extra("parse_error", &e.to_string());
                 return Err(HandlerErrorKind::InvalidHost(species, url.to_string()).into());
             }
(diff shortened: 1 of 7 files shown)


src/server/location.rs
```diff
@@ -265,7 +279,7 @@ impl Location {
             traits: None }
         }
         */
-        let mut result = LocationResult::default();
+        let mut result = self.default_loc.clone();
```
pjenvey, line 282: AFAICT this doesn't plug the defaults into `LocationResult` -- this only plugs it into the mmdb code.

src/settings.rs
```diff
@@ -106,11 +108,30 @@ impl Default for Settings {
...
+            return Err(ConfigError::Message("Missing adm_endpoint_url".to_owned()));
+        }
+        self.default_location = self.default_location.to_uppercase();
+        self.default_location.retain(|e| e.is_alphabetic());
+        if self.default_location.len() < 4 {
+            return Err(ConfigError::Message(
+                "Invalid default location specified. Please use a string like \"USOK\"".to_owned(),
+            ));
+        };
+        self.default_location = self.default_location[..4].to_owned();
```
pjenvey, line 128: if we're clipping the end to 4 I think we should call anything larger invalid.

src/settings.rs
```diff
@@ -80,6 +80,8 @@ pub struct Settings {
     pub test_file_path: String,
     /// Location test header override
     pub location_test_header: Option<String>,
+    /// Default location (if no location info is able to be determined for an IP)
+    pub default_location: String,
```
pjenvey, line 84: maybe "fallback location" describes this better than default?

pjenvey (changes requested):

src/settings.rs
pjenvey: Apparently 3166-2 (subdivision) is potentially 3 alphanumeric characters: 

https://en.wikipedia.org/wiki/ISO_3166-2

I was tempted to suggest the comma format but that doesn't matter. 

3166-1 alpha-2 (country) is two letter: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2

src/server/location.rs
jrconlin: Well, we're doing a bit of a shortcut here. I did see that we're not doing the same for the Google Header, so modified that. 

Also moved the location fixup code out of settings because it really should be part of the location impl.

src/adm/filter.rs
```diff
@@ -221,17 +221,17 @@ impl AdmFilter {
                 } else {
                     filter
                 };
-                if let Err(e) = self.check_advertiser(adv_filter, &mut tile, tags) {
+                if let Err(e) = self.check_advertiser(adv_filter, &mut tile, tags.clone()) {
```
pjenvey, line 224: this won't help as the `check_` methods end up with cloned tags but the `self.report` call handling the error doesn't use it (using the orig).

I'm fine reverting it back for now and fixing it later, I think it's more of an annoyance at this point

src/web/handlers.rs
```diff
@@ -53,7 +53,7 @@ pub async fn get_tiles(
     } else {
         Some(LocationResult::from_header(header, &state.settings))
     }
-    .unwrap_or_default();
+    .unwrap_or_else(|| LocationResult::from(&state.settings));
```
pjenvey, line 56: nit: I think both the mmdb and `from_header` clauses always return `Some()`, so this line's superfluous if the clauses just return their unwrapped results

src/server/location.rs
```diff
@@ -43,7 +56,7 @@ impl LocationResult {
             dbg!("Found Google Header");
             return Self::from_headervalue(header);
         }
-        Self::default()
+        Self::from(settings)
```
pjenvey, line 59: doh, we both missed this: It sounds like the LB's going to give us the header regardless of lookup failure, we just get empty values:

["The load balancer expands variables to empty strings when it cannot determine their values."](https://cloud.google.com/load-balancing/docs/custom-headers#variables)

pjenvey (changes requested):

src/adm/filter.rs
jrconlin: crap, yeah, what I get for being clever. I added the tag modifiers right before we do the error gen.

src/server/location.rs
jrconlin: ok, added a check for blank google headers. Also added a test for it.

src/web/handlers.rs
jrconlin: Yeah, I converted the Result to a `?`

src/server/location.rs
```diff
@@ -414,4 +442,32 @@ mod test {
...
+        assert!(settings.verify_settings().is_err());
+        settings.fallback_location = "Us, Oklahoma".to_owned();
+        assert!(settings.verify_settings().is_err());
+        settings.fallback_location = "us, Ok".to_owned();
+        assert!(settings.verify_settings().is_ok());
+        assert!(settings.fallback_location == *"USOK");
+
+        // From an empty Google LB header
+        let mut test_head = RequestHead::default();
+        let hv = "";
```
pjenvey, line 463: I think we're actually gonna get back ",":

https://github.com/mozilla-services/cloudops-infra/blob/master/projects/topsites/k8s/charts/contile/templates/backendconfig.yaml#L15

(probably just used solely "{client_region_subdivision}" which _seems_ to include both but I didn't dig deep enough into what "Unicode CLDR region code" meant by country)

jrconlin: I think we're good then. I altered the google header test locally to use `", "` and the code correctly returned the default.

jrconlin: let me push that here and we can keep it as a proof.

pjenvey: it's passing due to a nasty little bug. 

We get back "" for country and subdivision from `split()` so we have to check for those

src/server/location.rs
```diff
@@ -36,22 +49,25 @@ impl LocationResult {
...
             if let Some(header) = headers.get(loc_header) {
                 dbg!("Using test header");
-                return Self::from_headervalue(header);
+                return Self::from_headervalue(header, settings);
             }
         }
         if let Some(header) = headers.get(GOOG_LOC_HEADER) {
             dbg!("Found Google Header");
-            return Self::from_headervalue(header);
+            Self::from_headervalue(header, settings);
```
pjenvey, line 57: the nasty bug:
```suggestion
            return Self::from_headervalue(header, settings);
```

pjenvey (approved):

PR #127, opened 2021-05-28
bug: only permit Firefox UA from querying contile

### LINKED ISSUE #124: Contile dispatches a call to adM for non-Firefox user agents
Contile currently dispatches an API call to adM regardless of the user agent of the requester. Instead, Contile should skip making an API call to adM for clients with non-Firefox user agents and return a 403 status code to the client. We should consider adding a way to disable this functionality for testing/development purposes (something like a header or an environment variable).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-149)

jrconlin:
Closes: #124

## Description

Exclude non-firefox User Agents from getting results from Contile

## Testing

Tests included, but please try fetching tiles from `/v1/tiles` with various versions of firefox and other UAs to ensure that only firefox gets results. 

## Issue(s)

Closes #124

### DIFF
--- a/src/web/user_agent.rs
+++ b/src/web/user_agent.rs
@@ -2,10 +2,9 @@
 
 use std::fmt;
 
-use woothee::{
-    parser::{Parser, WootheeResult},
-    woothee::VALUE_UNKNOWN,
-};
+use woothee::parser::Parser;
+
+use crate::error::{HandlerErrorKind, HandlerResult};
 
 /// ADM required browser format form
 #[allow(dead_code)]
@@ -47,6 +46,7 @@ impl fmt::Display for OsFamily {
     }
 }
 
+/* Currently unused
 /// Strip a Firefox User-Agent string, returning a version only varying in Base
 /// OS (e.g. Mac, Windows, Linux) and Firefox major version number
 pub fn strip_ua(ua: &str) -> String {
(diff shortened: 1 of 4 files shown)


src/web/user_agent.rs
```diff
@@ -72,11 +72,18 @@ pub fn strip_ua(ua: &str) -> String {
...
 }
+*/
 
 /// Convert a UserAgent header into a simplified ([OsFamily], [FormFactor])
-pub fn get_device_info(ua: &str) -> (OsFamily, FormFactor) {
+pub fn get_device_info(ua: &str) -> HandlerResult<(OsFamily, FormFactor)> {
     let wresult = Parser::new().parse(ua).unwrap_or_default();
 
+    // If it's not firefox, it doesn't belong here...
+    // are there other names that could be returned? (e.g. for the iOS version)
```
pjenvey, line 82: Woothee only returns "Firefox" (also "FirefoxOS" which I doubt we'll hit any more).

We already have tests for iOS/android uas below that pass, so we can kill this comment I think

jrconlin: cool, wasn't sure if I was missing something.

pjenvey (approved):

PR #130, opened 2021-06-03
feat: add more metrics

### LINKED ISSUE #120: Emit a metric for location lookup failing
We forgot to add a metric for the case of location lookup failing #95.

This isn't critical at the moment but I think it's still worth adding at some point.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-147)

jrconlin:
Adds richer set of metrics and metric handling

Closes #120

## Description

Added more metric information for contile including:
* unknown locations
* adm tile issues

## Testing

Testing should result in increase of various metrics.

## Issue(s)

Closes #120

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/CONSVC-166)

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -244,19 +244,19 @@ impl AdmFilter {
                 };
                 if let Err(e) = self.check_advertiser(adv_filter, &mut tile, tags) {
                     dbg!("bad adv");
-                    metrics.incr_with_tags("tag.err.invalid_advertiser", Some(tags));
+                    metrics.incr_with_tags("filter.adm.err.invalid_advertiser", Some(tags));
                     self.report(&e, tags);
                     return None;
                 }
                 if let Err(e) = self.check_click(click_filter, &mut tile, tags) {
                     dbg!("bad click");
-                    metrics.incr_with_tags("tag.err.invalid_click", Some(tags));
+                    metrics.incr_with_tags("filter.adm.err.invalid_click", Some(tags));
                     self.report(&e, tags);
                     return None;
                 }
                 if let Err(e) = self.check_impression(impression_filter, &mut tile, tags) {
                     dbg!("bad imp");
-                    metrics.incr_with_tags("tag.err.invalid_impression", Some(tags));
+                    metrics.incr_with_tags("filter.adm.err.invalid_impression", Some(tags));
                     self.report(&e, tags);
                     return None;
                 }
@@ -269,7 +269,7 @@ impl AdmFilter {
                 ))
             }
             None => {
-                metrics.incr_with_tags("tag.err.unexpected_advertiser", Some(tags));
+                metrics.incr_with_tags("filter.adm.err.unexpected_advertiser", Some(tags));
                 self.report(
                     &HandlerErrorKind::UnexpectedAdvertiser(tile.name).into(),
                     tags,
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -10,6 +10,7 @@ use url::Url;
 use super::{tiles::AdmTile, AdmAdvertiserFilterSettings, Tile, DEFAULT};
 use crate::{
     error::{HandlerError, HandlerErrorKind, HandlerResult},
+    metrics::Metrics,
     tags::Tags,
     web::middleware::sentry as l_sentry,
 };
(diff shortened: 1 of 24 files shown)


src/server/location.rs
```diff
@@ -302,9 +305,19 @@ impl Location {
             Ok(location) => {
                 if let Some(names) = location.city.and_then(|c| c.names) {
                     result.city = get_preferred_language_element(&preferred_languages, &names)
+                } else {
+                    metrics.incr_with_tags(
+                        "location.unknown.city",
+                        Some(&Tags::from_extra(vec![("ip", ip_addr.to_string())])),
```
pjenvey, line 311: extras are only sent to sentry, so all these one offs in this module are unnecessary for metrics calls (I guess that applies to the new `from_extra` method too)

pjenvey (changes requested):
My other thought is we might want to prefix most of these metric names with `"adm."` -- it's always possible we plug in different partners later. Or does a tag make more sense?

I regret we missed prefixing the `CONTILE_{PARTNER_ID,SUB1}` settings

src/server/location.rs
jrconlin: Yeah, I debated adding these as proper tags, but I figured we'd hit cardinality issues pretty fast. I'll remove the tag from this for now. 

As for keeping the `from::extra`, I can see that potentially being useful in the future.

src/server/cache.rs
```diff
@@ -118,7 +138,13 @@ async fn tile_cache_updater(state: &ServerState) {
                 };
                 if new_tiles {
                     trace!("tile_cache_updater updating: {:?}", &key);
-                    tiles_cache.write().await.insert(key, Tiles { json: tiles });
+                    tiles_cache.write().await.insert(
+                        key,
+                        Tiles {
+                            json: tiles,
+                            ttl: SystemTime::now() + Duration::from_secs(settings.tiles_ttl as u64),
```
jrconlin, line 145: I'm debating if we should make this TTL * 2, to cover potential issues with ADM returning nothing, or an error. To be fair, we should probably also exit if adM returns anything other than a 200, and assume that they're having problems.

jrconlin: So, if a background refresh fails, how should we notify the client?

src/adm/filter.rs
```diff
@@ -238,16 +244,19 @@ impl AdmFilter {
                 };
                 if let Err(e) = self.check_advertiser(adv_filter, &mut tile, tags) {
                     dbg!("bad adv");
+                    metrics.incr_with_tags("tag.err.invalid_advertiser", Some(tags));
```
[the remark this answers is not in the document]
jrconlin: ah, I think that was the product of a bad replace.
I'll correct the names (and I'll also add `adm`)

src/server/location.rs
```diff
@@ -302,9 +304,13 @@ impl Location {
             Ok(location) => {
                 if let Some(names) = location.city.and_then(|c| c.names) {
                     result.city = get_preferred_language_element(&preferred_languages, &names)
+                } else {
+                    metrics.incr("location.unknown.city");
                 };
                 if let Some(names) = location.country.and_then(|c| c.names) {
                     result.country = get_preferred_language_element(&preferred_languages, &names)
+                } else {
```
pjenvey, line 312: We could add these (country/subdivision) for the location header cases too

src/server/cache.rs
```diff
@@ -71,14 +78,25 @@ async fn tile_cache_updater(state: &ServerState) {
...
         reqwest_client,
         adm_endpoint_url,
-        metrics,
+        settings,
         ..
     } = state;
 
     trace!("tile_cache_updater..");
-    let keys: Vec<_> = tiles_cache.read().await.keys().cloned().collect();
+    let tiles = tiles_cache.read().await;
```
pjenvey, line 86: I'm a little concerned that the cache read lock (guard) is now held throughout this function and across await points, we should probably avoid that if we can.

jrconlin: Moved the cleanup routine into it's own block and have the locks more isolated.

src/server/location.rs
jrconlin: Added the 'unknown' metrics to the google header code.

pjenvey (approved):

PR #133, opened 2021-06-04
feat: redirect root requests to official documentation

### LINKED ISSUE #131: Add handler for / to direct to documentation or information page
Per recommendation in bug https://bugzilla.mozilla.org/show_bug.cgi?id=1576919, we should add a handler to direct users who visit the server root / to a page that contains documentation or information about this service.

This issue is blocked until that documentation is available.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-167)

jrconlin:
Closes: #131

## Description

Redirects root GET requests to official documentation URL

## Testing

`curl -v http://{host}` should return a 303 with a `location` header set to the `settings.documentation_url`

## Issue(s)

Closes #131

┆Issue is synchronized with this [Jira Bug](https://jira.mozilla.com/browse/CONSVC-170)

### DIFF
--- a/src/web/dockerflow.rs
+++ b/src/web/dockerflow.rs
@@ -11,6 +11,7 @@ use actix_web::{web, HttpRequest, HttpResponse};
 use serde_json::Value;
 
 use crate::error::HandlerError;
+use crate::server::ServerState;
 
 /// Well Known DockerFlow commands for Ops callbacks
 pub const DOCKER_FLOW_ENDPOINTS: [&str; 4] = [
@@ -26,7 +27,8 @@ pub fn service(config: &mut web::ServiceConfig) {
         .service(web::resource("/__lbheartbeat__").route(web::get().to(lbheartbeat)))
         .service(web::resource("/__heartbeat__").route(web::get().to(heartbeat)))
         .service(web::resource("/__version__").route(web::get().to(version)))
-        .service(web::resource("/__error__").route(web::get().to(test_error)));
+        .service(web::resource("/__error__").route(web::get().to(test_error)))
+        .service(web::resource("").route(web::get().to(document_boot)));
 }
 
 /// Used by the load balancer to indicate that the server can respond to
@@ -60,4 +62,14 @@ async fn test_error(_: HttpRequest) -> Result<HttpResponse, HandlerError> {
     // generate an error for sentry.
     error!("Test Error");
     Err(HandlerError::internal("Oh Noes!"))
+}
+
+async fn document_boot(
+    _: HttpRequest,
+    state: web::Data<ServerState>,
+) -> Result<HttpResponse, HandlerError> {
+    let settings = &state.settings;
+    return Ok(HttpResponse::SeeOther()
+        .header("Location", settings.documentation_url.clone())
+        .finish());
 }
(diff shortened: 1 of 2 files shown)


mythmon:
> curl -v http://{host} should return a 303 with a location header set to the settings.documentation_url

In Merino, I opted for 302 Found instead of 303 See Other. I wasn't really confident about that though. We should probably use the same code on all our services. Why did you choose 303 here?

src/web/dockerflow.rs
```diff
@@ -61,3 +63,13 @@ async fn test_error(_: HttpRequest) -> Result<HttpResponse, HandlerError> {
     error!("Test Error");
     Err(HandlerError::internal("Oh Noes!"))
 }
+
+async fn document_boot(
+    _: HttpRequest,
```
pjenvey, line 68: nit: `HttpRequest`'s just another extractor, might as well kill this line since it isn't used anyway (I think I removed the others in the skeleton)

pjenvey (commented):
> In Merino, I opted for 302 Found instead of 303 See Other.

+1 for 302 since this is a GET, looks like 303's for `PUT/POST`s redirecting to `GET`s.

jrconlin:
I chose 303 here mostly because while the URL is valid, the preferred content resides elsewhere. I've always considered 301/302 to be more "This URL is no longer valid, go *here* from now on."

mythmon:
I specifically avoided 301 because it can have some unfortunate caching side effects that I've found to be far longer lived than I liked in the past. 302's temporary nature was appealing to me. How is 303 handled by browser caches?

jrconlin:
I don't believe that 303 is cached by browsers, at least [not that I've been able to find](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#targets_of_caching_operations).

src/web/dockerflow.rs
jrconlin: I removed the `HttpRequest` noop from the other calls.

mythmon:
Quoting the RFCs:

[302](https://datatracker.ietf.org/doc/html/rfc7231#section-6.4.3):

> indicates that the target resource resides temporarily under a different URI

[303](https://datatracker.ietf.org/doc/html/rfc7231#section-6.4.4):

> indicates that the server is redirecting the user agent to a different resource

The main question here seems to be: is the page we are redirecting to the requested resource? I think it is: bug 1576919 asked for

> if a browser were to hit the root of the service, say "https://[service-name].update-services.mozilla.org/" that they be redirected to a Wikimo page describing what that service does and why.

That is, the root of the service should be a page describing what the service does and why. The page we are redirecting to *is* that documentation, not a different resource.

I could also seen an argument that requests to `/` are asking for a different thing, but I don't know what that thing could be.

jrconlin:
Yeah, I get the feeling that it's probably a bike shed, so not going to be super zealot about any of this.

I consider the other dockerflow endpoint stuff (e.g. `__heartbeat__`, `__version__`) to be valid resources hanging off of the root, so `/` is still a valid endpoint. Just more a case of the princess being in another castle. 

I know that 303 is supported by [all the browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303#browser_compatibility), so the naive user who just types in "http://contile.services.mozilla.com/" into their browser doesn't know or care what the mechanics are, so much as they care where they go and what they get at the end of the process. 

Ultimately, I don't really care enough to fight about this. If 302 is what other services are using, awesome.

pjenvey (approved):

PR #140, opened 2021-06-07
feat: add timeout for calls to ADM service

### LINKED ISSUE #139: Add timeout for ADM requests
Currently we use the default reqwest timeout for ADM tile fetches. We should be more explicit about what timeout we want to use.

jrconlin:
## Description

* Uses setting `adm_timeout`, and defaults to 5 seconds
I suspect that we probably may have different timeouts for various
partners. There is no timeout specified for a default reqwest builder.

## Testing

requests to ADM should take no longer than 5 seconds in default setting

## Issue(s)

Closes #139

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -58,6 +58,8 @@ pub struct Settings {
     pub adm_max_tiles: u8,
     /// number of tiles to query from ADM (default: 10)
     pub adm_query_tile_count: u8,
+    /// Timeout requests to the ADM server after this many seconds (default: 3)
+    pub adm_timeout: u64,
     /// Expire tiles after this many seconds (15 * 60s)
     pub tiles_ttl: u32,
     /// ADM tile settings (either as JSON or a path to a JSON file)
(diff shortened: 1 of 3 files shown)


src/settings.rs
```diff
@@ -58,6 +58,8 @@ pub struct Settings {
     pub adm_max_tiles: u8,
     /// number of tiles to query from ADM (default: 10)
     pub adm_query_tile_count: u8,
+    /// Timeout requests to the ADM server after this many seconds (default: 3)
```
pjenvey, line 61:
```suggestion
    /// Timeout requests to the ADM server after this many seconds (default: 5)
```

pjenvey (changes requested):

src/settings.rs
jrconlin: heh, oh, you saw that too? 🤦🏻‍♂️

pjenvey (approved):

PR #144, opened 2021-06-08
Feat/123 ignore

### LINKED ISSUE #123: Allow setting an advertiser "ignore"
We're reporting "unknown" advertisers as an event to sentry (#50) for a couple reasons.

We may want a setting for explicitly "ignoring" certain advertisers to no longer report this entry to sentry. Particularly as every pod currently acts independently, logging each event on their own (resulting in duplicates), cluttering the Sentry dashboard.

E.g.:

- Sentry reports a new advertiser of "Acme"
- Ad Ops decides we do not want it included
- Set "Acme" for ignore so the events are no longer generated

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-148)

jrconlin:
Closes #123

## Description

Adds `adm_ignore_advertisers` optional JSON formatted list of ignored advertisers

**Note** this is a non-breaking change currently, but the following arguments are now considered OBSOLETE and will generate a warning. Expect that these values will be dropped in a future version.

* `sub1`  | `CONTILE_SUB1` => `adm_sub1` | `CONTILE_ADM_SUB1`
* `partner_id` | `CONTILE_PARTNER_ID` => `adm_partner_id` | `CONTILE_PARTNER_ID`

## Testing

unit test included.

## Issue(s)

Closes #123

### DIFF
--- a/src/adm/settings.rs
+++ b/src/adm/settings.rs
@@ -1,9 +1,17 @@
-use std::{collections::HashMap, fmt::Debug, fs::File, path::Path};
+use std::{
+    collections::{HashMap, HashSet},
+    fmt::Debug,
+    fs::File,
+    path::Path,
+};
 
 use serde::{ser::SerializeSeq, Deserialize, Deserializer, Serialize, Serializer};
 
 use super::AdmFilter;
-use crate::{error::HandlerResult, settings::Settings};
+use crate::{
+    error::{HandlerError, HandlerResult},
+    settings::Settings,
+};
 
 /// The name of the "Default" node, which is used as a fall back if no data
 /// is defined for a given partner.
@@ -40,6 +48,7 @@ pub struct AdmAdvertiserFilterSettings {
     /// Optional set of valid regions for the tile (e.g ["en", "en-US/TX"])
     #[serde(default)]
     pub(crate) include_regions: Vec<String>,
+    pub(crate) ignore_advertisers: Option<Vec<String>>,
 }
 
 /// Parse JSON:
@@ -131,8 +140,18 @@ impl From<&Settings> for HandlerResult<AdmFilter> {
             // map the settings to the URL we're going to be checking
             filter_map.insert(adv.to_lowercase(), setting);
         }
+        let ignore_list: HashSet<String> = serde_json::from_str(
+            &settings
+                .adm_ignore_advertisers
+                .clone()
+                .unwrap_or_else(|| "[]".to_owned()),
+        )
+        .map_err(|e| {
+            HandlerError::internal(&format!("Invalid ADM Ignore list specification: {:?}", e))
+        })?;
         Ok(AdmFilter {
             filter_set: filter_map,
+            ignore_list,
         })
     }
 }
--- a/src/adm/settings.rs
+++ b/src/adm/settings.rs
@@ -93,8 +93,31 @@ pub(crate) type AdmSettings = HashMap<String, AdmAdvertiserFilterSettings>;
 /// This allows `CONTILE_ADM_SETTINGS` to either be specified as inline JSON, or if the
 /// Settings are too large to fit into an ENV string, specified in a path to where the
 /// settings more comfortably fit.
-impl From<&Settings> for AdmSettings {
-    fn from(settings: &Settings) -> Self {
+impl From<&mut Settings> for AdmSettings {
+    fn from(settings: &mut Settings) -> Self {
+        // TODO: Convert these to macros.
+        if settings.adm_sub1.is_none() {
+            if settings.sub1.is_some() {
+                settings.adm_sub1 = settings.sub1.clone();
+                warn!(
+                    "`{}` is obsolete and will be removed. Please use `{}`",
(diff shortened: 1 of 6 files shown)


src/settings.rs
```diff
@@ -99,22 +108,26 @@ impl Default for Settings {
...
-            partner_id: "demofeed".to_owned(),
-            sub1: "123456789".to_owned(),
             test_mode: false,
             test_file_path: "./tools/test/test_data/".to_owned(),
             location_test_header: None,
             fallback_location: "USOK".to_owned(),
             documentation_url: "https://developer.mozilla.org/".to_owned(),
+            // ADM specific settings
+            adm_endpoint_url: "".to_owned(),
+            adm_partner_id: Some("demofeed".to_owned()),
```
pjenvey, line 121: this and adm_sub1 need to be `None` for the fallback to work

src/adm/settings.rs
```diff
@@ -84,8 +93,31 @@ pub(crate) type AdmSettings = HashMap<String, AdmAdvertiserFilterSettings>;
...
 /// settings more comfortably fit.
-impl From<&Settings> for AdmSettings {
-    fn from(settings: &Settings) -> Self {
+impl From<&mut Settings> for AdmSettings {
+    fn from(settings: &mut Settings) -> Self {
+        // TODO: Convert these to macros.
+        if settings.adm_sub1.is_none() {
+            if settings.sub1.is_some() {
+                settings.adm_sub1 = settings.sub1.clone();
+                warn!(
```
pjenvey, line 102: Let's switch the `warn!`s to `eprintln!`s -- so they always show up regardless of log levels and more importantly logging isn't initialized yet when this is called, so they'll get lost.

pjenvey (changes requested):

src/settings.rs
pjenvey: The last commit switch these back to a `Some`

pjenvey (changes requested):
^ Some vs None

src/settings.rs
jrconlin: Yes, because otherwise an error would be produced. Perhaps I don't quite understand your original concern. Both the new and old parameters need to be optional otherwise the application will fail if they're not present. This will break older settings files as well as newer ones that have been switched to use the new names. The application code looks to see if the new values have been specified, and if not, back fills them using the older values with a warning. The unit test has been updated to show that the proper behavior is performed.

If these were set to None, then no default value was specified for these and these values would need to be set in things like the unit tests, which is a different behavior than what was originally done in the app. 

Was there a different concern you had in mind?

pjenvey: The issue with them being `Some` is it means the `eprintln!` warnings are never triggered as they're guarded in e.g. `settings.adm_sub1.is_none()` -- which is never true when there's a `Some`.

pjenvey: Also the legacy settings become overwritten by the `Some`s -- e.g. `CONTILE_PARTNER_ID="foobar" CONTILE_SUB1="22" cargo run` ends up with "demofeed"/"123456789" instead for the new settings which `get_tiles` ends up using.

jrconlin: I'm still not sure I understand the problem.

Yes, they're guarded against. For the warning to trigger, you need to not have `adm_sub1` set, but have `sub1` set, with the presumption that there will be a cross over period where both could be set and we want to have values for the new options for unit tests. There is no warning generated if you're using the new argument name, since it's value is used while the old one is ignored if present. 

When the old arguments are removed from the code, things will behave as they would if you either specified an unknown env var or an invalid option.

Do you want us to continue to accept and flag the obsolete argument once it's no longer accepted?

jrconlin: Ok, updated the tests to capture that. thanks!

src/adm/settings.rs
```diff
@@ -123,16 +155,56 @@ impl From<&Settings> for AdmSettings {
...
+        let ignore_list = settings
+            .adm_ignore_advertisers
+            .clone()
+            .unwrap_or_else(|| "[]".to_owned());
         for (adv, setting) in AdmSettings::from(settings) {
             trace!("Processing records for {:?}", &adv);
             // map the settings to the URL we're going to be checking
             filter_map.insert(adv.to_lowercase(), setting);
         }
+        let ignore_list: HashSet<String> = serde_json::from_str(&ignore_list).map_err(|e| {
```
pjenvey, line 170: We forgot one last thing: these need to be lowercase'd. Which suggests it could use a quick test too if it's not too much trouble.

pjenvey (changes requested):

src/adm/settings.rs
jrconlin: Not sure how to best test this, since the ignore list is basically doing the same thing as an invalid, just not reporting things out to Sentry. v. 0.19 doesn't include the inspection hacks. I'll add a test that verifies that the ignore list is lowercased for now.

pjenvey (approved):

PR #146, opened 2021-06-08
feat: add an optional trace header to metrics/logging

### LINKED ISSUE #145: Add trace context to metric and logging calls
Per Ops, we should add the trace header data to the metric and logging calls.

This header may be `X-Cloud-Trace-Context`, ops to confirm.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-180)

jrconlin:
## Description

This adds a `trace_header` setting. This will add the value (if found in
the Headers) to the tags as `header.trace`.

(e.g. if a request has `X-Cloud-Trace-Context: abc123` this will add a
tag with `header.trace: abc123` to the outboud tags for logging and
metrics.)
## Testing

Ensure the content of the tracing header is sent to metric and logging.

## Issue(s)

Closes #145

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -2,11 +2,16 @@
 
 use std::collections::HashMap;
 
+use actix_web::{web::Data, HttpRequest, dev::ServiceRequest};
 use config::{Config, ConfigError, Environment, File};
 use serde::Deserialize;
 
 use crate::adm::AdmSettings;
-use crate::server::{img_storage::StorageSettings, location::Location};
+use crate::server::{
+    img_storage::StorageSettings,
+    location::Location,
+    ServerState,
+};
 
 static PREFIX: &str = "contile";
 
@@ -86,6 +91,8 @@ pub struct Settings {
     pub fallback_location: String,
     /// URL to the official documentation
     pub documentation_url: String,
+    /// Operational trace header
+    pub trace_header: Option<String>,
 }
 
 impl Default for Settings {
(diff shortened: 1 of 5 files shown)


src/settings.rs
```diff
@@ -204,6 +208,23 @@ impl Settings {
     }
 }
 
+// TODO: someone better at lifetimes could fix these.
+// The returned settings need only last as long as the request.
+// we clone to break out of the Arc
+impl From<&HttpRequest> for Settings {
```
pjenvey, line 214: This was trickier than I thought it was gonna be 😅 but I think this works:

https://github.com/mozilla-services/contile/compare/feat/145-trace...feat/145-trace-lifetimes

Wanna apply the same to `impl From<&ServiceRequest> for Settings` below?

jrconlin: Oh, cool! Thanks!
Yep, I'll modify the other ones along the same lines.

jbuck:
@jrconlin sorry, I don't understand this code so I'm not sure where to add my comment... but it's critical that the trace tag _isn't_ added to statsd, and only added to sentry/log output. The trace id is a random string which will blow up influxdb cardinality

jrconlin:
@jbuck yeah, that's exactly the sort of comment I need. thanks!

src/tags.rs
```diff
@@ -127,6 +129,15 @@ impl From<&RequestHead> for Tags {
                 extra.insert("ua".to_owned(), uas.to_string());
             }
         }
+        if let Some(tracer) = settings.trace_header.clone() {
```
pjenvey, line 132: nit

```suggestion
        if let Some(ref tracer) = settings.trace_header {
```

pjenvey (approved):

pjenvey (approved):

PR #150, opened 2021-06-09
feat: Return "NaN" if no region is found for a location

### LINKED ISSUE #148: Don't fallback on unknown subdivisions
We're seeing clients with a country determined by the LB but not a subdivision, which we fallback to "OK", causing bogus adM parameters of e.g. `country-code=FR&subdivision=OK`.

Instead let's pass adM an empty subdivision if one cannot be determined, their service seems ok with this instead of a bogus value.

We likely still want a fallback country (US) in case it could not be determined.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-197)

jrconlin:
Closes #148

## Description

Return the string "NaN" if no region information could be determined for a location lookup.

## Testing

Unit test included

## Issue(s)

Closes #148.

### DIFF
--- a/src/server/location.rs
+++ b/src/server/location.rs
@@ -35,7 +35,11 @@ impl From<&Settings> for LocationResult {
         let default_loc = settings.fallback_location.clone();
         Self {
             city: None,
-            subdivision: Some(default_loc[2..4].to_string()),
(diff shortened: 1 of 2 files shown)


pjenvey:
superseded by #156

PR #152, opened 2021-06-10
feat: integrate cache updating into the GET request

### LINKED ISSUE #151: Defer cache updating to get_tiles handler
Our initial impl of the cache handles updating (when tiles expire) via the periodic reporter, running on an interval of `tiles_ttl`. Every \<TTL\> seconds it updates each cache key sequentially.

This has potential issues with a larger number of cache keys: e.g. potentially delayed responses from the adM side could cause the periodic job to take a very long time (potentially even longer than \<TTL\> seconds). Additionally the partner requested we avoid this technique (though the reasoning behind that request is up for debate).

Let's move the expiry logic into the GET request. This has its own potential issues: e.g. multiple clients triggering a large number of concurrent refreshes upon expiration (resulting in many simultaneous network connections and potentially cache writes) but these are typical problems for a proxy with potential mitigations if needed.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-195)

pjenvey:
## Description

avoiding potential issues w/ the periodic reporter managing it and large cache
sizes

also switch RwLock<HashMap> to the concurrent DashMap for its performance
benefits

## Issue(s)

Closes #151


Cargo.toml
```diff
@@ -24,6 +24,7 @@ cadence = "0.25"
 docopt = "1.1"
 cloud-storage = "0.6" # 0.7+ includes request 0.11, tokio 1.4
 config = "0.11"
+dashmap = "4.0.2"
```
pjenvey, line 27: one note about dashmap: it's susceptible to a [theoretical spinlock deadlock situation](https://github.com/xacrimon/dashmap/issues/142).

I'm not very worried about this and there's already a [PR in progress](https://github.com/xacrimon/dashmap/pull/145) to fix it.

src/web/handlers.rs
```diff
@@ -98,47 +100,53 @@ pub async fn get_tiles(
             None
         },
     )
-    .await
-    {
+    .await;
+
+    let json = match result {
         Ok(response) => {
-            // adM sometimes returns an invalid response. We don't want to cache that.
```
pjenvey, line 104: Note: these are valid responses from adM so we're now caching them

src/server/mod.rs
```diff
@@ -88,24 +88,22 @@ impl Server {
     /// initialize a new instance of the server from [Settings]
     pub async fn with_settings(settings: Settings) -> Result<dev::Server, HandlerError> {
         let filter = HandlerResult::<AdmFilter>::from(&settings)?;
+        let metrics = metrics_from_opts(&settings)?;
+        let tiles_cache = cache::TilesCache::new(384);
```
jrconlin, line 92: 384 Seems a bit specific. What's the source of this? Should it be a setting?

jrconlin (approved):

src/server/mod.rs
pjenvey: I looked at the size of the cache yesterday and it was =~ 350, so just something bigger than the original 75. Then as of this morning it's more like =~ 550.

It's not terribly important (just preferably something larger than the default of 8 or whatever it is) so I don't think it warrants a setting.

jrconlin: Fair.  Maybe add a comment to that effect like
```
// This value is arbitrary and based on the expected mean. Feel free to adjust.
```
Or make it a `const` so that it's just a bit less magical feeling. I can see this becoming a legendary value for stupid reasons.

jrconlin (dismissed):

jrconlin (approved):

PR #153, opened 2021-06-10
bug: make UA name an 'extra' value for Sentry errors.

### LINKED ISSUE #147: Convert Invalid UA to metric
https://sentry.prod.mozaws.net/operations/contile/issues/11582997/?environment=prod&referrer=alert_email

We will sometimes see spiders and other UA, because of course we will. We should convert these to metrics.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-181)

jrconlin:
## Description
In order to reduce some of the cardinality we're seeing with some sentry errors, it may be a good idea to move some of the data as "tag" info. (e.g. for InvalidUA and ADM Server Errors, we could move the "unique" bits as extra tag info so that we get a better idea of how many events we're seeing without using regular expression tricks.

## Testing
test included (for InvalidUA at least)

## Issue(s)

Closes #147

### DIFF
--- a/src/adm/tiles.rs
+++ b/src/adm/tiles.rs
@@ -164,7 +164,9 @@ pub async fn get_tiles(
             .await
             .map_err(|e| {
                 // ADM servers are down, or improperly configured
-                HandlerErrorKind::AdmServerError(e.to_string())
+                let mut err: HandlerError = HandlerErrorKind::AdmServerError().into();
+                err.tags.add_extra("error", &e.to_string());
+                err
             })?
             .error_for_status()?
             .json()
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -69,7 +69,9 @@ impl AdmFilter {
     fn report(&self, error: &HandlerError, tags: &Tags) {
         // trace!(&error, &tags);
         // TODO: if not error.is_reportable, just add to metrics.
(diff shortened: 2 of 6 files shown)


src/adm/tiles.rs
```diff
@@ -164,7 +164,9 @@ pub async fn get_tiles(
             .await
             .map_err(|e| {
                 // ADM servers are down, or improperly configured
-                HandlerErrorKind::AdmServerError(e.to_string())
+                let mut err: HandlerError = HandlerErrorKind::AdmServerError().into();
```
pjenvey, line 167: We might want to keep the string for now and instead break down the type of error (later). Most of what we're seeing so far is [err.is_timeout()](https://docs.rs/reqwest/0.11.3/reqwest/struct.Error.html#method.is_timeout), so we'd probably start there -- but probably be handy to know if it's something else before we fully break it all down.

pjenvey: Oh but the string's obviously sent as an extra (I'm not sure I noticed this 🤷‍♂️ ), so we're not losing it, this works for me.

We could map specific e.g. `err.is_timeout()` errors to their own individual error if we wanted later.

pjenvey (approved):

Issue #149, opened 2021-06-09
Wire in Maxmind as the primary location lookup source


pjenvey:
With us already seeing situations like #148 it makes sense to try out maxmind in case it results in less failed location lookups.

Let's finish wiring this into the app and use maxmind as the primary location lookup source, while still falling back to the LB header if it fails.

Let's ping SRE (@jbuck) on how to integrate the commercial version into contile.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/CONSVC-196)

jrconlin:
The code is already wired up to use the Max Mind database (if there is one defined in the `maxminddb_loc` setting), and will fail back to looking for the LB header if a lookup produces an error. We do not attempt to back fill from the Header and would need to add code if that is desired.

jbuck:
https://github.com/mozilla-services/cloudops-infra/pull/3193

pjenvey:
This need some [additional work](https://github.com/mozilla-services/contile/pull/155#issuecomment-861857026) to follow up #155

PR #154, opened 2021-06-14
feat: include ip addr for diagnosing mmdb lookup failures

pjenvey:
## Testing

Changes https://sentry.prod.mozaws.net/operations/contile/issues/11643624/?environment=dev

## Issue(s)

Issue #149

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -44,7 +44,12 @@ pub async fn get_tiles(
         let addr = match ip_addr_str.parse() {
             Ok(v) => v,
             Err(e) => {
-                return Err(HandlerErrorKind::General(format!("Invalid remote IP {:?}", e)).into());
+                // Temporary: log the IP addr for debugging mmdb issues
+                return Err(HandlerErrorKind::General(format!(
+                    "Invalid remote IP ({:?}) {:?}",
+                    ip_addr_str, e
+                ))
+                .into());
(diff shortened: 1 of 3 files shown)


src/web/handlers.rs
```diff
@@ -44,7 +44,12 @@ pub async fn get_tiles(
         let addr = match ip_addr_str.parse() {
             Ok(v) => v,
             Err(e) => {
-                return Err(HandlerErrorKind::General(format!("Invalid remote IP {:?}", e)).into());
+                // Temporary: log the IP addr for debugging mmdb issues
+                return Err(HandlerErrorKind::General(format!(
```
mythmon, line 48: Since IPs are relatively sensitive information, should this be a configurable option that defaults to off? We could then turn it back off without having to cut a new release? (Or is there some other plan that isn't linked to from here?)

mythmon (changes requested):

jbuck (approved):

src/web/handlers.rs
pjenvey: I can but I'm not intending for this to be used outside of the dev env.

I'm working on a change for #148 that will land shortly where I'll be intending to revert his

mythmon (commented):
The code changes look good. I trust that we'll be careful to not let this ride to prod.

PR #155, opened 2021-06-14
feat: get the client IP for mmdb from X-Forwarded-For

pjenvey:
## Testing

I'm working on a separate PR that refactors location lookup into a separate extractor, then we can more easily add unit tests for this.

## Issue(s)

Issue #149

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -35,6 +35,8 @@ pub async fn get_tiles(
     let mut addr = None;
     if let Some(header) = request.headers().get(&*X_FORWARDED_FOR) {
         if let Ok(value) = header.to_str() {
+            // Expect a typical X-Forwarded-For where the first address is the
+            // client's, the front ends should ensure this
             addr = value
                 .split(',')
                 .next()
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -1,5 +1,8 @@
 //! API Handlers
-use actix_web::{web, HttpRequest, HttpResponse};
+use std::net::{IpAddr, SocketAddr};
+
+use actix_web::{http::HeaderName, web, HttpRequest, HttpResponse};
+use lazy_static::lazy_static;
 
 use super::user_agent;
 use crate::{
@@ -10,6 +13,10 @@ use crate::{
     tags::Tags,
     web::{extractors::TilesRequest, middleware::sentry as l_sentry},
 };
+
+lazy_static! {
+    static ref X_FORWARDED_FOR: HeaderName = HeaderName::from_static("x-forwarded-for");
+}
 
 /// Handler for `.../v1/tiles` endpoint
 ///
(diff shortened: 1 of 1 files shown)


src/web/handlers.rs
```diff
@@ -24,34 +31,31 @@ pub async fn get_tiles(
...
-        let default = state
-            .adm_country_ip_map
-            .get("US")
-            .expect("Invalid ADM_COUNTRY_IP_MAP settting");
-        if let Some(country) = &treq.country {
-            state.adm_country_ip_map.get(country).unwrap_or(default)
-        } else {
-            default
+    let mut addr = None;
+    if let Some(header) = request.headers().get(&*X_FORWARDED_FOR) {
```
jrconlin, line 36: Well, crap. https://docs.rs/actix-web/3.3.2/actix_web/dev/struct.ConnectionInfo.html#method.realip_remote_addr

pjenvey: I based some of this code off of that method -- problem with it is its us of `Forwarded` makes it problematic (it's possibly spoofed unless nginx/LB are disallowing it)

jbuck: Docs on how it works: https://cloud.google.com/load-balancing/docs/https#x-forwarded-for_header

The way nginx parses the client IP out of the X-Forwarded-For header is something you'll want to copy - basically you put a bunch of trusted IPs, and the first untrusted IP it finds is what is considered the client IP address: http://nginx.org/en/docs/http/ngx_http_realip_module.html

jrconlin: @jbuck, I think I remember a lot of this, and I think I had to deal with it with ChannelServer. I can't remember if we need to toss the trusted IPs or if they're trimmed before the header is handed off.

jbuck: They're not trimmed, here's an example from the logs: `"99.237.130.164, 34.96.87.143, 99.237.130.164"` (also there's a bug we need to fix where the the load balancer IP is repeated)

jrconlin: Ah, right: https://github.com/mozilla-services/channelserver/blob/fc2c592f832740e340a5b901ed9c87d56d4ce015/channelserver/src/meta.rs#L166

ncloudioj: @jbuck Looks like we are not using the `Real IP` Nginx module for [Contile](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/topsites/k8s/charts/contile/templates/nginx-configmap.yaml). Shall we add those to the configuration file?

```
real_ip_header X-Forwarded-For;
set_real_ip_from 0.0.0.0/0;
# set_real_ip_from other_cdn_ips;
```

src/web/handlers.rs
```diff
@@ -24,34 +31,31 @@ pub async fn get_tiles(
...
+        if let Ok(value) = header.to_str() {
+            addr = value
+                .split(',')
+                .next()
```
ncloudioj, line 40: Nit: let's add a comment to clarify that the real client IP should be the first chunk of `X-Forwarded-For` in our system setup (GCLB -> Openresty (Real IP) -> Contile).

src/web/handlers.rs
```diff
@@ -65,7 +69,6 @@ pub async fn get_tiles(
     {
         tags.add_extra("country", &location.country());
         tags.add_extra("region", &location.region());
```
ncloudioj, line 71: Just curious, is it possible that `LocationResult::from_header(header, settings, &metrics)` could return incorrect values, like those reported by adM last week? If that's the case, we'd need to handle that so that no incorrect values get sent over.

ncloudioj (changes requested):

src/web/handlers.rs
pjenvey: Yea, that's a separate issue #148 that I'll have a separate PR for

src/web/handlers.rs
pjenvey: 👍  w/ more restrictive subnets than 0.0.0.0/0 and `real_ip_recursive on` (copy over topsites proxy's nginx config here, it looks good)

ncloudioj (approved):
LGTM. Thanks!

jbuck:
This PR needs some additional fixes for the case when a malicious user is sending additional `X-Forwarded-For` headers. Check this example:

```
# My IP address
$ curl checkip.amazonaws.com
99.237.130.164

# GCLB IP address
$ dig +short contile-dev.topsites.nonprod.cloudops.mozgcp.net
34.96.87.143

# Sending spoofed IP address
curl -H 'X-Forwarded-For: fakeip' https://contile-dev.topsites.nonprod.cloudops.mozgcp.net/__version__
{"commit":"2a3882de79f1b91d55ad0d910b69ee2aeea480dc","version":"","source":"https://github.com/mozilla-services/contile","build":"https://circleci.com/gh/mozilla-services/contile/852"}

# Results in nginx
x_forwarded_for: "fakeip, 99.237.130.164, 34.96.87.143"
```

This PR's code takes the first address as the client IP address, but instead it should reverse the split array, and return the first untrusted IP address. In this example we'd trust `34.96.87.143/32` but we'd need to specify multiple IP ranges. If you wanted to specify an array index from the end of the list that'd work too.

ncloudioj:
> This PR's code takes the first address as the client IP address, but instead it should reverse the split array, and return the first untrusted IP address. In this example we'd trust 34.96.87.143/32 but we'd need to specify multiple IP ranges. If you wanted to specify an array index from the end of the list that'd work too.

To simplify things, I think it's fine to stick to the existing first-entry extraction strategy. If people want to mess around with us by either using the fake IP or the invalid CIDRs, as long as Contile can properly handle it without crashing. It will just fall back to GCLB headers, and then fall back to the default country.

@pjenvey @jbuck What do you think?

pjenvey:
@ncloudioj I'm fine w/ releasing with this current impl. so we can start testing/rolling out the recent location changes. We can log a separate issue for tightening this up and aim to get it out shortly in a separate dot release. I'll defer this to @jbuck though

PR #156, opened 2021-06-15
feat: don't fallback on unknown subdivisions

### LINKED ISSUE #148: Don't fallback on unknown subdivisions

pjenvey:
## Description

only fallback on country (US) when no location info can be found

and recognize mmdb's now always available in tests

## Issue(s)

Closes #148

### DIFF
--- a/src/server/location.rs
+++ b/src/server/location.rs
@@ -32,11 +32,10 @@ pub struct LocationResult {
 
 impl From<&Settings> for LocationResult {
     fn from(settings: &Settings) -> Self {
-        let default_loc = settings.fallback_location.clone();
         Self {
             city: None,
-            subdivision: Some(default_loc[2..4].to_string()),
-            country: Some(default_loc[..2].to_string()),
+            subdivision: None,
+            country: Some(settings.fallback_country.clone()),
             dma: None,
         }
     }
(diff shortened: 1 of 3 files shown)


src/server/location.rs
```diff
@@ -240,16 +239,13 @@ impl Location {
...
 
-    pub fn fix(location_str: &str) -> Result<String, ConfigError> {
-        let mut loc = location_str.to_uppercase();
-        loc.retain(|e| e.is_alphabetic());
-        let llen = loc.len();
-        if !(4..=5).contains(&llen) {
+    pub fn fix_fallback_country(fallback_country: &str) -> Result<String, ConfigError> {
+        if fallback_country.len() != 2 {
             return Err(ConfigError::Message(
-                "Invalid default location specified. Please use a string like \"USOK\"".to_owned(),
```
ncloudioj, line 249: Ah, this is where "OK" is coming from ;)

ncloudioj (approved):
:shipit:

PR #164, opened 2021-06-21
remove temporary archival note

pjenvey:


ncloudioj (approved):
Oops, looks like Clippy was not happy with some recent changes.

PR #199, opened 2021-06-24
fix: return 204 no contents for cache hits 

### LINKED ISSUE #191: Return 204 No Content for cache hits
#152 began caching empty result sets but upon a cache hit we always return 200 OK.

Cache hits should check for empty tiles and return 204s in the case of empty tiles.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1095)

pjenvey:
## Testing

new integration test passes

## Issue(s)

Closes #191

### DIFF
--- a/src/server/cache.rs
+++ b/src/server/cache.rs
@@ -9,7 +9,7 @@ use std::{
 use cadence::StatsdClient;
 use dashmap::DashMap;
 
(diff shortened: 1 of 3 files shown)


ncloudioj (approved):
Looks great, thanks!

PR #212, opened 2021-07-08
feat: Send image URLs to CDN.

### LINKED ISSUE #167: Tile Images to CDN (Contile image hosting)
Store images provided in the adM Instant Suggest tiles to our own Content Distribution Network (CDN) system. This is the top priority for Contile v2.

Business need: Updates to existing built-in advertiser icons have to ride the train, which is too long of a delay if an icon needs to be updated for any reason. The ability to update images off-train is also a key value prop of Contile.

Privacy need: This would increase user privacy by potentially not leaking information to an unknown CDN.

[https://github.com/mozilla-services/contile/issues/19](https://github.com/mozilla-services/contile/issues/19)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-33)
┆Epic: Contile v2

jrconlin:
Closes #167

## Description

Send ADM provided images to a CDN.

This will activate the `img_storage` mod, and the `storage` JSON config option. 

## Testing

TODO: Add unit tests for 
* config options
* CDN image load
* validate that img URLs are getting swapped appropriately

## Issue(s)

Closes #167


src/server/img_storage.rs
```diff
@@ -228,6 +250,9 @@ impl StoreImage {
             .await
             .map_err(|e| HandlerErrorKind::Internal(format!("Image body error: {:?}", e)))?;
 
+        let meta = self.meta(&image)?;
```
ncloudioj, line 253: For context, per newtab specs, images for TopSites should be square, if `heigh != width`, we can treat it as an invalid image. Also, the minimal hi-res icon size for TopSites is 96 pixels, so if it's below that, we can also treat it as invalid.

Assuming we're going to use `image_url` for the URL, shall we call it `image_size` for this one?

jrconlin: Works for me! Thanks!

are there any other things we should check the image for? (e.g. max size?, unusual format? (not "jpg" or "png") etc.)

ncloudioj: I think we can enforce the following things here:
* Size: `width == height && 96 <= wide <= 256`
* MIME extensions: Firefox only takes any of `png, jpg, svg`

ncloudioj:
@jrconlin Thanks for working on this feature!

Just wanted to add a quick note that when we upload the image to GCS, let's make sure the following headers are attached as well

```
{
    'Cache-Control': 'public, max-age=31536000', // cache for 1 year
    'Content-Disposition': 'inline' // show this image inline on the page other than showing a "Save as" prompt
}
```

jrconlin:
I know that the CDN uses the `Cache-Control` header we set for the bucket, but not sure how to set the `Content-Disposition` header.

ncloudioj:
Looks like `Content-Disposition` is supported by its [XML API](https://cloud.google.com/storage/docs/xml-api/reference-headers). IIRC, `inline` is the default value for `Content-Disposition`, if our library doesn't support it, let's just keep an eye on the browser making sure it behaves properly.

src/adm/tiles.rs
```diff
@@ -87,20 +87,27 @@ pub struct Tile {
...
 impl Tile {
-    pub fn from_adm_tile(tile: AdmTile, position: Option<u8>) -> Self {
+    pub fn from_adm_tile(tile: AdmTile, position: Option<u8>, new: bool) -> Self {
         Self {
             id: tile.id,
             name: tile.name,
             url: tile.advertiser_url,
             click_url: tile.click_url,
             image_url: tile.image_url,
+            image_size: None,
```
pjenvey, line 106: Is this field still needed when `ImageMetrics` has a size?

We _could_ also cleanup this response output but it's not terribly important (nor do I know how it would affect the client impl cc: @ncloudioj)

e.g. 
```json
  "click_url": "...",
  image: {
    "url": "...",
    "size": "...",
    "width": "etc",
  }
```

src/server/img_storage.rs
```diff
@@ -54,35 +121,72 @@ pub struct StoreResult {
     pub url: uri::Uri,
     pub hash: String,
```
pjenvey, line 122: Might wanna also kill these unused fields (like `hash`) until they're needed.

Another general nit -- I've been trying not to include "Result" in struct names such as this one as it makes them sound like they're a variant of a `Result` (such as `HandlerResult`) to fresh eyes. Maybe `StoredImage`

pjenvey (changes requested):

src/server/img_storage.rs
```diff
@@ -1,21 +1,77 @@
...
+    bucket_ttl: u64,
+    /// The max time to live for cached data, ~ 15 days.
+    #[serde(default = "default_ttl")]
+    cache_ttl: u64,
+    /// Max dimensions for an image
+    #[serde(default = "ImageMetricSettings::default")]
+    metrics: ImageMetricSettings,
+}
+
+fn default_ttl() -> u64 {
```
pjenvey, line 68: nit: since StorageSettings has a `Default` impl we shouldn't need these functions and their associated `#[serde(default = "default_ttl")]`.

Instead we can use the `#[serde(default)]` container attribute on the entire struct (see above) to fill these in from the `Default` impl.

src/adm/tiles.rs
jrconlin: Part of the confusion is that "size" apparently is different. The UA wants "size" to be the width/height. We want "size" to be the length of the image (to construct the file name).

ncloudioj: Firefox only expects the `image_url` and the `image_width` from Contile. By historical reason, we reference `image_width` as `image_size` in the Fx codebase.

Alternatively, We can also leave out `image_width` from the response payload if Contile can enforce this invariant:
Contile would only provide the `image_url` field in the response if `image.width == image.height &&  96 <= image.width <= 256`.

So when we see `image_url` in the response, we know that the image is of the valid format, we can simply assign the minimal favicon width to it, and Firefox knows how to render it properly.

@jrconlin Let me know which one you prefer so that I can adjust my client patch accordingly.

jrconlin: I'll add this as a comment to the code and only return `image_url` and `image_size`.

ncloudioj: sounds good!

pjenvey: > Alternatively, We can also leave out `image_width` from the response payload if Contile can enforce this invariant:
> Contile would only provide the `image_url` field in the response if `image.width == image.height && 96 <= image.width <= 256`.

@jrconlin It looks like we're doing exactly this, so it sounds like we could only return `image_url`?

jrconlin: I'd like to return both, since the size could vary between 96 and 128 px.
(I'll also say that previous experience has taught me to return more info than I need in these cases, but I'll defer to the folk doing the work.)

ncloudioj: Agreed, let's also return `image_size` other than letting UA hardcode it with `MIN_FAVICON_SIZE (96)`.

BTW, Dao and Amy have reviewed the client-side [patch](https://phabricator.services.mozilla.com/D119923), I am landing it to nightly now.

src/server/img_storage.rs
```diff
@@ -160,19 +248,31 @@ impl StoreImage {
...
+    pub fn as_hash(&self, source: &Bytes) -> String {
+        let mut hasher = DefaultHasher::new();
+        source.hash(&mut hasher);
+        base64::encode_config(hasher.finish().to_ne_bytes(), base64::URL_SAFE_NO_PAD)
+    }
+
+    /// Fetch the bytes for an image based on a URI
+    pub(crate) async fn fetch(&self, uri: &uri::Uri) -> HandlerResult<(Bytes, String)> {
+        trace!("fetching... {:?}", &uri);
         let res = reqwest::get(&uri.to_string())
```
pjenvey, line 267: - Let's give `StorageImage` a reference to the `reqwest_client` built at `Server` creation and call reqwest through that instead (to take advantage of the underlying per host pooling)
- Probably worth adding a timeout to the get call like we do for adM
- Definitely want an `.error_for_status()?` for the get too
- Reqwest errors have an `Into<HandlerError>` so map_err shouldn't be necessary (the reqwest error and its traceback should tell us enough about this error)

src/server/img_storage.rs
```diff
@@ -199,11 +393,21 @@ impl StoreImage {
...
-                hash: v.etag.clone(),
-                url: public_url.parse()?,
-                object: v,
-            }),
+            Ok(mut object) => {
+                object.content_disposition = Some("inline".to_owned());
+                object.cache_control = Some(format!("public, max-age={}", self.settings.cache_ttl));
+                object.update().await.map_err(|_| {
+                    error!("Could not set disposition for {:?}", object.self_link);
+                    HandlerErrorKind::BadImage("Could not set content disposition")
```
pjenvey, line 401: Nothing to hold this up over but I think some of the Internal and BadImage errors in here derived from `cloud_storage::Error` should have their own `HandlerKind` `CloudStorage(#[from] cloud_storage::Error)` variant

src/server/img_storage.rs
```diff
@@ -214,3 +418,133 @@ impl StoreImage {
...
+        let src_img = "https://evilonastick.com/test/128px.jpg";
+
+        let test_settings = test_storage_settings();
+        let bucket = StoreImage::build_bucket(&test_settings)
+            .await
+            .unwrap()
+            .unwrap();
+        let target = src_img.parse::<Uri>().unwrap();
+        let result = bucket.store(&target).await;
+        assert!(result.is_ok());
```
pjenvey, line 500: debugging nit:

This assert ends up hiding the error, this'll show it:

```suggestion
        bucket.store(&target).await.expect("Store failed");
```

src/server/img_storage.rs
```diff
@@ -181,16 +281,110 @@ impl StoreImage {
...
+            return Err(HandlerError::internal("No storage bucket defined"));
+        }
 
-        // store data to the googles
+        // image paths tend to be "https://<host>/account/###/###/####.jpg"
+        // They may be unreliable as a hash source, so use the image bytes.
+        let image_path = format!(
+            "{}.{}.{}",
+            self.as_hash(&image),
+            image.len(),
```
pjenvey, line 366: :smart:

pjenvey (changes requested):
I'm still looking at the filename hashing bits but this is looking good

src/server/img_storage.rs
jrconlin: Yeah, I figure we can revisit and revise the errors later, probably when we break things up to componentize stuff.  For now, having the one, semi-universal errors is "good enough".

pjenvey (approved):

pjenvey:
r+ sorry for the delay

PR #214, opened 2021-07-17
test(Integration_test): add integration test to the CircleCI workflow

ashrivastava-qa:
## Description
`Automated the process of running the integration tests to trigger the CircleCI workflow. Below are the changes made:`
`- integrated CircleCI workflow with the integration tests in the Contile repo to trigger the workflow`
`- added a `docker-compose.yml` file with the image details for Partner, Client and Contile`
`- added `/volumes` folder where the test `scenarios.yml`, `responses.yml` and `adm_settings.json` files resides required to run the tests`
`- updated the CircleCI `config.yml` file with the necessary steps to run the integration tests with commands`

## Testing

@mozilla-services/context-services - Please review the files added together with the commands. Thanks!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -230,10 +230,6 @@ workflows:
           name: contile-integration-tests
           requires:
           - build
-          filters:
-            branches:
-              only:
-                - main
       - deploy:
           requires:
             - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,24 @@ jobs:
       - rust-check
       - rust-clippy
 
-  build-and-test:
+  test:
+    docker:
+      - image: circleci/rust:latest
+        auth:
+          username: $DOCKER_USER
+          password: $DOCKER_PASS
+        environment:
+          RUST_BACKTRACE: 1
+          # XXX: begin_test_transaction doesn't play nice over threaded tests
+          RUST_TEST_THREADS: 1
+    steps:
+      - checkout
+      - setup-rust
+      - write-version
+      - cargo-build
+      - run-tests
+
+  build:
     docker:
       - image: circleci/rust:latest
         auth:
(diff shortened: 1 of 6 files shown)


jrconlin (approved):
LGTM. (Wondering if I should remove the "Fake-Response" / test mode flags from the tile code)

.circleci/config.yml
```diff
@@ -186,13 +216,28 @@ workflows:
...
               only: /.*/
+          requires:
+            - build
+      - contile-integration-tests:
+          name: contile-integration-tests
+          requires:
+          - build
+          filters:
+            branches:
+              only:
```
pjenvey, line 235: Sounds like this won't add much time to CI so let's run this everywhere like the other targets. It'll be nice to have it catch issues on pull requests.

pjenvey (changes requested):

jrconlin:
I don't think we need it currently, but I wonder if we should use something like [`store_test_results`](https://circleci.com/docs/2.0/configuration-reference/#storetestresults) if we're running tests like these? It may be overkill.

.circleci/config.yml
ashrivastava-qa: okay. Made the changes. Thanks for the r+!

jrconlin (approved):

pjenvey (approved):

jrconlin:
By the way, THANK YOU for doing this! This is a huge help.

pjenvey:
> I don't think we need it currently, but I wonder if we should use something like [`store_test_results`](https://circleci.com/docs/2.0/configuration-reference/#storetestresults) if we're running tests like these? It may be overkill.

That does look like a nice feature (particularly the test insights):

https://circleci.com/docs/2.0/collect-test-data/

We'd need the integration test's pytest runner to output `--junitxml=out.xml`. Also the docker compose would probably need a volume mount letting it write out that `out.xml`

pjenvey:
Oh but the Insights is for the self hosted CircleCI Server only though

jrconlin:
Yeah, I think it just needs the workspace to compile the report, but I'm not sure. 

Something to play around with in the future, I suppose.

ashrivastava-qa:
Yeah, I think that's a good suggestion @jrconlin and @pjenvey. I'll try and see if I can find out more about it to be implemented. Thanks! :)

PR #221, opened 2021-07-23
Feat/201 jitter

### LINKED ISSUE #201: Add TTL jitter
We should consider adding a small amount of jitter to tiles' expiration time. The goal would be to discourage the cache from triggering refresh requests to partners all at the same time.

E.g. a plus or minus percentage of overall time seconds to the 15 minute TTL.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1101)
┆Epic: Contile v2

jrconlin:
## Description

Adds a setting to add "jitter" to a given tiles in memory cache TTL

## Testing

Set the `CONTILE_JITTER=0` and note that tiles  do not display jitter and expire exactly at `CONTILE_TTL`.
Set the `CONTILE_JITTER=50` and note that tiles expire at any time between 50 and 150% of `CONTILE_TTL`.

## Issue(s)

Closes #201


src/web/handlers.rs
```diff
@@ -1,15 +1,28 @@
...
 };
 
+/// Calculate the ttl from the settings by taking the tiles_ttl
+/// and calculating a jitter that is no more than 50% of the total TTL.
+/// It is recommended that "jitter" be 10%.
+pub fn add_jitter(settings: &Settings) -> u32 {
+    let mut rng = thread_rng();
+    let ftl = settings.tiles_ttl as f32;
+    let offset = ftl * (std::cmp::min(settings.jitter, 50) as f32 * 0.01);
+    let jit = rng.gen_range(0.0 - offset..offset);
```
pjenvey, line 22: Shouldn't it be inclusive (I don't actually care 😄 )?

```suggestion
    let jit = rng.gen_range(0.0 - offset..=offset);
```

pjenvey (approved):

PR #223, opened 2021-07-26
feat: record metric for empty ADM responses

### LINKED ISSUE #222: Log empty results
We're seeing a higher number of empty return responses than we expected. We should record an error to sentry when we return an empty result to provide more information about why that's happening. (e.g. empty response from ADM, request error, etc.)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1157)

jrconlin:
Closes #222

## Description

Record a metric when either ADM fails to return any tiles `filter.adm.empty_response`, or if no tiles pass filtering `filter.adm.all_filtered`)

## Testing

Generate a fake response that either returns no data or returns data that does not pass filtering. Monitor outbound metric record.

## Issue(s)

Closes #222


src/web/handlers.rs
```diff
@@ -110,14 +110,11 @@ pub async fn get_tiles(
     }
 }
 
-fn content_response(content: &cache::TilesContent, metrics: &Metrics) -> HttpResponse {
+fn content_response(content: &cache::TilesContent, _metrics: &Metrics) -> HttpResponse {
```
pjenvey, line 113: let's just kill this arg

src/web/user_agent.rs
```diff
@@ -50,6 +50,12 @@ pub struct DeviceInfo {
     pub os_family: OsFamily,
 }
 
+impl ToString for DeviceInfo {
```
pjenvey, line 53: `ToString` docs recommend you don't implement it directly, impl `Display` instead.

However I don't think we need it or the new `Location::as_string` at all since this just adds new metrics calls where extras aren't used.

If we did want these extras included in later occurring errors this probably isn't the best place to add them. I don't think they're needed as adM errors like [this one](https://sentry.prod.mozaws.net/operations/contile/issues/11738905/?query=is%3Aunresolved) include the adM url in the error string with that information (we could add that adM url explicitly as an extra but `reqwest::Error` seems to always include it for us anyway)

pjenvey (approved):

PR #234, opened 2021-08-03
feat: Add local server hostname to tags

### LINKED ISSUE #200: Include a hostname tag in metrics
So we can group some metrics by the k8s pod let's include the hostname in metrics as a tag. We probably only need the tag for certain metrics, most importantly the ones emitted by the cache's periodic reporter

jrconlin:
## Description

Adds the local server hostname to the metric and logging tags. 

## Testing

Unit test modified to include check for presence. 

## Issue(s)

Closes #200

### DIFF
--- a/src/tags.rs
+++ b/src/tags.rs
@@ -78,8 +78,12 @@ pub fn parse_user_agent(agent: &str) -> (WootheeResult<'_>, &str) {
 /// `extra` are not searchable, but may not be sent to [crate::metrics::Metrics].
 #[derive(Clone, Debug)]
 pub struct Tags {
+    // All tags (both metric and sentry)
     pub tags: HashMap<String, String>,
+    // Sentry only "extra" data.
     pub extra: HashMap<String, String>,
+    // metric only supplemental tags.
+    pub metric: HashMap<String, String>,
 }
 
 impl Default for Tags {
@@ -87,6 +91,7 @@ impl Default for Tags {
         Tags {
             tags: HashMap::new(),
             extra: HashMap::new(),
+            metric: HashMap::new(),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/tags.rs
```diff
@@ -126,6 +126,13 @@ impl Tags {
                 insert_if_not_empty("ua.os.family", metrics_os, &mut tags);
                 insert_if_not_empty("ua.os.ver", &ua_result.os_version.to_owned(), &mut tags);
                 insert_if_not_empty("ua.browser.ver", ua_result.version, &mut tags);
+                insert_if_not_empty(
```
pjenvey, line 129: We also want this new tag emitted by the periodic reporter metrics, unfortunately it does not get any tags via `Tags::from_head` (since it's outside of a web request). Like in syncstorage, we probably just need a one off call to gethostname in that code path.

It's too bad we call `gethostname` every request here but I guess it's cheap and currently a bit difficult to do once up front.

pjenvey (dismissed):

jbuck:
I am concerned this tag will be too high granularity - every single k8s pod will have a different hostname. What are you trying to measure by adding the hostname?

jrconlin:
per the source issue:
```So we can group some metrics by the k8s pod let's include the hostname in metrics as a tag. ```

We can probably skip adding this tag to anything other than metric reporting, but we'll probably still hit granularity issues.

jrconlin:
Ok, refactored a bit.
I added a `metric` section to `Tags` that includes tags that only go to metrics (they're mixed in right before sending them out via cadence). That should prevent the tag from going out to Sentry by accident, as well as make it very clear when adding what's happening. 

I also added the tag only in `get_tiles`, which is only when we get the tiles. FWIW, also renamed the old periodic reporter, since it's less about re-fetching everything, and more about doing garbage collection.

pjenvey:
> I am concerned this tag will be too high granularity - every single k8s pod will have a different hostname. What are you trying to measure by adding the hostname?

Nothing too specific but it's helpful for diagnosing general individual pod behavior, allowing more correlation across metrics. E.g. the most recently spun up pod has this certain CPU usage with this number of adM request calls.

Though this is a good point, one I hadn't considered because we emit hostname tags for all autopush and some syncstorage metrics. I suppose it's a potential issue there but nobody's complained (yet).

Would a hostname tag in practice be much different from e.g. k8s metrics' `resource.label.pod_name`?

jrconlin:
huh. So, looking at: https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/ I don't believe that pods generally know what their name is, that info would have to be exposed in some way. 

I'm going to drop this one in favor of better identifying the requirement in #200

jrconlin:
In light of the conversation in #200, reopening.

jbuck (approved):

PR #219, opened 2021-07-21
refactor: Replace location determination with common-rs crate

mythmon:
## Description

The location parsing code that Contile had was useful, so I stole it to put into https://github.com/mozilla-services/common-rs/pull/14. This PR re-integrates that code as an external crate. Note that this is currently pulling the crate straight from that PR. That means the other PR should merge and this PR should be updated to point to crates.io before we merge this PR.

This drops support for GCS provided location headers, which are not being used currently. We should consider adding that feature to the shared crate as well.

## Testing

MaxMind based location should work exactly the same as before. Non-location features should be unaffected.

PR #235, opened 2021-08-03
feat: enable send of dma-code

### LINKED ISSUE #205: Provide targeting by DMA
We should provide position targeting by DMA in addition to the existing Location lookup data. 

DMA can be retrieved from the MaxMind geoip lookup database.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1110)
┆Epic: Contile v2

jrconlin:
Closes #205

## Description

Sends `dma-code` data to partner

## Testing

Notice inclusion of `dma-code` parameter in data sent to partner.

## Issue(s)

Closes #205


hackebrot:
I created https://github.com/mozilla-services/contile-integration-tests/pull/63 to add the `dma-code` query parameter to the partner API in the integration tests.

src/server/cache.rs
```diff
@@ -18,7 +18,9 @@ pub struct AudienceKey {
     /// Country in ISO 3166-1 alpha-2 format
     pub country_code: String,
     /// Region/subdivision (e.g. a US state) in ISO 3166-2 format
-    pub region_code: String,
+    pub region_code: Option<String>,
+    /// The DMA code (u16)
+    pub dma_code: Option<u16>,
```
pjenvey, line 23: confirmed, this should indeed be included in the cache key

ncloudioj:
adM's will add dma to their CDN cache as well. They will let us know once that's ready on their side. We can merge it after that.

src/server/location.rs
```diff
@@ -124,12 +124,25 @@ impl LocationResult {
         headers.get(GOOG_LOC_HEADER).is_some()
     }
 
-    pub fn region(&self) -> String {
-        self.subdivision.clone().unwrap_or_default()
+    pub fn region(&self) -> Option<String> {
+        if self.country() == "US" {
```
pjenvey, line 128: Wasn't the conclusion here is we'll continue doing what we're doing, always send region regardless of country (and they'll begin caching it outside of US and correct this in their API docs)? cc @ncloudioj

src/server/location.rs
```diff
@@ -124,12 +124,25 @@ impl LocationResult {
...
+        }
+        None
     }
 
     pub fn country(&self) -> String {
-        self.country.clone().unwrap_or_default()
+        self.country.clone().unwrap_or_default().to_uppercase()
+    }
+
+    pub fn dma(&self) -> String {
```
pjenvey, line 138: nit: consider keeping this as u16 to match [actix-web-location](https://github.com/mozilla-services/contile/pull/219) which we'll switch to shortly

src/server/location.rs
jrconlin: Oh, I thought it was the opposite. I'll defer to you and Nan's decision here.

src/server/location.rs
jrconlin: I might make this an `Option<u16>` so that the "None" case is still captured.

src/server/location.rs
jrconlin: ah, OBE.

ncloudioj:
@jrconlin Confirmed with security & legal that we wanted to enforce that "do not send if its population is less than 15k".

I did some digging, looks like we have three DMA codes meeting that criterion:

```
dma-code | county
----------------------
552 	Presque Isle
583 	Alpena
798 	Glendive
```

Could you add a DMA filter for this, please?

src/adm/tiles.rs
```diff
@@ -113,6 +113,20 @@ impl Tile {
     }
 }
 
+pub fn filtered_dma(settings: &Settings, dma: &u16) -> String {
+    if settings
+        .excluded_dma
+        .clone()
```
ncloudioj, line 119: Nit: I think you could do this to save a clone.
```rs
if settings
    .excluded_dma
    .as_ref()
    .unwrap_or(&vec![])
    .contains(dma)
```

jrconlin: I keep forgetting about `.as_ref()`. Thanks!

src/web/handlers.rs
```diff
@@ -60,9 +60,16 @@ pub async fn get_tiles(
         // tags.clone().commit(&mut request.extensions_mut());
     }
 
+    // Only set the region if the country is US
+    // see "Instant Find | API Response Caching" in the 7_jul_21 spec.
```
ncloudioj, line 64: I think adM wanted us to keep the current use of "region-code" for non-US countries in the API request and the cache key. I am confirming this with them again now, let's see what they say.

src/settings.rs
```diff
@@ -81,6 +81,9 @@ pub struct Settings {
     pub adm_ignore_advertisers: Option<String>,
     /// a JSON list of advertisers to allow for versions of firefox less than 91.
     pub adm_has_legacy_image: Option<String>,
+    /// a JSON list of location DMAs to never return (population less than 15K)
+    pub exclude_dma: Option<String>,
+    pub excluded_dma: Option<Vec<u16>>,
```
pjenvey, line 86: I think this should live in `ServerState` instead, otherwise this adds an ignored `CONTILE_EXCLUDED_DMA` setting that could be confused for the real setting later.

It'd just need the deserializing moved into `Server::with_settings`

src/web/handlers.rs
ncloudioj: Confirmed that we should continue sending/caching "region-code" for all the countries. adM will update their API doc to reflect this.

ncloudioj (approved):
LGTM, r=nanj

adM is ready to take dma-code from us, feel free to merge it so that we can deploy it next time.

Thanks!

PR #236, opened 2021-08-04
fix: Write version.json before docker build

jbuck:
In bc8d5f29e406f57cfbbae533ee5cbd21cd49e38e the step to write version.json was moved from build to test. This meant that the default version.json in the repository was written to the docker image. Because the default version.json doesn't have a `BUILD_URL` defined, Jenkins has not been deploying contile

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,7 +111,6 @@ jobs:
     steps:
       - checkout
       - setup-rust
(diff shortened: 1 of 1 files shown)


ncloudioj (dismissed):
:shipit:

jrconlin (approved):

PR #247, opened 2021-08-06
feat: Add `adm_has_legacy_image` setting to filter <v91 tiles

### LINKED ISSUE #246: Filter outbound advertisers based on UA version number
We will need to limit the campaigns we display based on the User Agent version. This is because 91+ allows for pulling images from CDN where -91 can only use internal images. 

This should be a temporary fix, and noted as such. 

This will require creating a list of limited campaign advertisers to allow for versions of Firefox less than 91. Any ADM provided advertiser that is not contained in that list should be dropped from the set of tiles returned to the UA. 

Coordination with QA regarding integration testing will be required.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1181)
┆Epic: Contile v2

jrconlin:
This adds `adm_has_legacy_image` which is a JSON list of advertisers
that have "legacy" images pre-loaded into firefox.

Closes #246

## Description

Describe these changes.

## Testing

How should reviewers test?

## Issue(s)

Closes [link](link).

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -18,7 +18,7 @@ use crate::{
     server::location::LocationResult,
     tags::Tags,
     web::middleware::sentry as l_sentry,
-    web::DeviceInfo,
+    web::{DeviceInfo, OsFamily},
 };
 
 lazy_static! {
@@ -262,9 +262,13 @@ impl AdmFilter {
                     metrics.incr_with_tags("filter.adm.err.invalid_location", Some(tags));
                     return None;
                 }
+                // match to the version that we switched over from built in image management
+                // to CDN image fetch. Note: iOS does not use the standard firefox version number
 
-                if device_info.ff_version < 91
-                    && !self.legacy_list.contains(&tile.name.to_lowercase())
+                if device_info.os_family == OsFamily::IOs && device_info.ff_version < 36
+                    || device_info.os_family != OsFamily::IOs
+                        && device_info.ff_version < 91
+                        && !self.legacy_list.contains(&tile.name.to_lowercase())
                 {
                     trace!("Rejecting tile: Not a legacy advertiser {:?}", &tile.name);
                     metrics.incr_with_tags("filter.adm.err.non_legacy", Some(tags));
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -18,6 +18,7 @@ use crate::{
     server::location::LocationResult,
     tags::Tags,
     web::middleware::sentry as l_sentry,
+    web::DeviceInfo,
 };
 
 lazy_static! {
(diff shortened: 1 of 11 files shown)


src/adm/filter.rs
```diff
@@ -258,6 +263,14 @@ impl AdmFilter {
                     return None;
                 }
 
+                if device_info.ff_version < 91
+                    && !self.legacy_list.contains(&tile.name.to_lowercase())
```
pjenvey: The legacy status of the requesting browser should also be included in the cache key (`AudienceKey`) since it affects the Tiles content.

pjenvey (changes requested):

src/adm/filter.rs
ncloudioj: Good catch. This would add a multiplier to the current cache keyspace.

Shall we add the status to the cache key (multiplier: 2), or add the ff_version (multiplier: N)?

pjenvey: I think a boolean flag of legacy or not is all we need/want.

jrconlin: Way ahead of y'all. I added a function to `DeviceInfo` to return if we should provide legacy only. 

I was debating if the cache should store the first pass validated ADM result rather than the finalized list to provide a pool of data we can progressively filter later, but that's YAGNI.

ncloudioj (approved):
Thanks!

src/web/user_agent.rs
```diff
@@ -48,6 +49,15 @@ impl fmt::Display for OsFamily {
 pub struct DeviceInfo {
     pub form_factor: FormFactor,
     pub os_family: OsFamily,
+    // We only care about major versions.
+    pub ff_version: u32,
+}
+
+impl DeviceInfo {
+    pub fn legacy_only(&self) -> bool {
+        self.os_family == OsFamily::IOs && self.ff_version < 36
```
pjenvey, line 58: We're serving Desktop only, was there a reason for checking less than < 36 on iOS? (I guess it doesn't matter much either)

src/web/test.rs
```diff
@@ -192,6 +195,46 @@ async fn basic() {
...
+        content_type
+            .unwrap()
+            .to_str()
+            .expect("Couldn't parse Content-Type"),
+        "application/json"
+    );
+
+    let result: Value = test::read_body_json(resp).await;
+    let tiles = result["tiles"].as_array().expect("!tiles.is_array()");
+    assert!(tiles.len() > 1);
```
pjenvey, line 229: Probably just forgot to modify this from `basic` -- we should assert_eq len() of 2 and that "Dunder Mifflin" was filtered out

pjenvey (changes requested):

src/web/user_agent.rs
jrconlin: The fact that both versions differ was only noted when I updated the UA strings for testing. I'm fine saying desktop only for this check but I'd like to prevent potential future issues. In this case "You Hope You Ain't Gonna Need It".

pjenvey (approved):

pjenvey (approved):

PR #250, opened 2021-08-09
feat: reduce redundant adM requests

### LINKED ISSUE #248: Reduce concurrent adM requests
When first populating or refreshing contile's cache with content from adM, multiple requests for the same content can be triggered simultaneously. Many of these request are redundant, which:

1. add extra work for contile to process and write them to the cache (see #46 for potential cache side effects)
1. cause extra traffic to adM (particularly problematic during downtime conditions #220)

With Contile now handling image hosting its processing task is potentially lengthier, which could result in even larger number of redundant concurrent requests triggered (since the cache doesn't fill as quickly).

To improve this: when an adM request is triggered Contile could write this event to its cache, letting subsequent requests act more intelligently. Instead of spawning more requests they could instead: wait for that adM request to finish (by simply 503ing the client or actually waiting over an spmc channel for its completion) or by simply returning stale content until the request finishes.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1184)

pjenvey:
## Description

by storing state about active adM requests in the cache

when a request's in the process of populating a cache entry for the first time,
subsequent requests will 503 until it's completed

when a request's in the process of refreshing an existing cache entry,
subsequent requests will return the stale entry until it's completed

and cache empty for BadAdmResponse errors (which haven't happened yet)

## Issue(s)

Closes #248

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -6,7 +6,11 @@ use crate::{
     adm,
     error::{HandlerError, HandlerErrorKind},
     metrics::Metrics,
-    server::{cache, location::LocationResult, ServerState},
+    server::{
+        cache::{self, Tiles, TilesState},
+        location::LocationResult,
+        ServerState,
+    },
     settings::Settings,
     tags::Tags,
     web::{middleware::sentry as l_sentry, DeviceInfo},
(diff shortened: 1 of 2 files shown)


src/web/handlers.rs
```diff
@@ -88,36 +136,75 @@ pub async fn get_tiles(
...
+                    _ => Err(e),
                 }
-                _ => Err(e),
             }
         }
+    };
+
+    let result = handle_result();
+    // Cleanup the TilesState on errors
+    // TODO: potential panics are not currently cleaned up
```
pjenvey, line 187: I'm hoping we could move this cleanup section into a `Drop` impl of some new struct to handle this. The struct could probably also handle inserting/altering the new state values above and hopefully simplfy the handler a bit

src/web/handlers.rs
```diff
@@ -61,18 +65,62 @@ pub async fn get_tiles(
...
+        // First make a cheap read from the cache
+        if let Some(tiles_state) = state.tiles_cache.get(&audience_key) {
+            match &*tiles_state {
+                TilesState::Populating => {
+                    // Another task is currently populating this entry and will
+                    // complete shortly. 503 until then instead of queueing
+                    // more redundant requests
+                    trace!("get_tiles: Another task Populating");
+                    metrics.incr("tiles_cache.miss.populating");
+                    return Ok(HttpResponse::ServiceUnavailable().finish());
```
pjenvey, line 80: I've begun some work on making this case instead "watch" for completion via a tokio watch channel. I'm not yet sure it's worth it (if this will even be that significant number of 503s)

ncloudioj: I was gonna comment on this `return-503-if-populating`. If we end up seeing a high volume of 503s, Firefox might miss some impressions&clicks. We might want to let those requests "spin" a bit in hope that the population could be done soon.

Without any instrumentation, I am inclined to treat that as a potential follow-up before further complicating this patch.

ncloudioj (approved):
LGTM. Thanks!

jbuck:
I'm not in love with the idea of sending a 503 to the client while the cache entry is being created - a series of 503's is the thing that's going to page me... guess it depends how many it'll cause

jrconlin:
If I understand the flow correctly, a 503 would be returned if the cache is empty or that if there's no matching cached value and one thread is in the process of creating that tile entry. I can see this producing a spike of 503s if:
* one or more nodes are brought on-line during a high traffic period
* ADM servers are "lagging" for any period while a node is coming on-line

does that sound correct @pjenvey ?

pjenvey:
> I'm not in love with the idea of sending a 503 to the client while the cache entry is being created - a series of 503's is the thing that's going to page me... guess it depends how many it'll cause

Totally, as Nan mentioned we'll revisit this depending on the load testing/prod results.

I went w/ the 503s for now because

- I didn't see very many 503s in local load testing even with a 2 second delay added to the adm fetching
- This should at least produce less 5xx than the current release's last load test
- this patch complicates the handler enough as is (and there's probably some room for improvement)

pjenvey:
> If I understand the flow correctly, a 503 would be returned if the cache is empty or that if there's no matching cached value and one thread is in the process of creating that tile entry. I can see this producing a spike of 503s if:

Basically yes (though if the entry is empty a fetch is triggered on that request).

> 
>     * one or more nodes are brought on-line during a high traffic period
> 
>     * ADM servers are "lagging" for any period while a node is coming on-line

And they can only lag so much on us since we enforce a timeout (though we may be missing one other form of time outs  #253)

ncloudioj:
I'd recommend landing it as it is and then keeping a close eye on the load tests and prod metrics. If it does generate a significant amount of 503s, we still have a few options for improvement, such as returning 204 or applying a sleep-and-retry.

PR #260, opened 2021-08-12
feat: add a cloud storage write precondition

### LINKED ISSUE #259: Upload images with an ifGenerationMatch 0 parameter
To avoid redundant, concurrent writes to the image cloud storage bucket (which can trigger rate limit errors), we should be able to limit the writes by specifying a precondition to match the first generation of the object.

So once the first write to the object completes, subsequent attempts should fail the precondition.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1205)

pjenvey:
## Description

so we only issue writes when the object doesn't already exist, avoiding rate
limiting errors

This supersedes the bits of #258 adding the precondition check

## Testing

Please run the img_storage tests with a test bucket setup

## Issue(s)

Closes #259

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ bytes = "1.0"
 cadence = "0.26"
 chrono = "0.4"
 docopt = "1.1"
-cloud-storage = { git = "https://github.com/mozilla-services/cloud-storage-rs", branch = "feat/create_with_params-0.6.2" } # 0.7+ includes request 0.11, tokio 1.4
+cloud-storage = { git = "https://github.com/mozilla-services/cloud-storage-rs", branch = "release/0.6.2-create_with_params" } # 0.7+ includes request 0.11, tokio 1.4
 config = "0.11"
 dashmap = "4.0.2"
 futures = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ bytes = "1.0"
 cadence = "0.26"
 chrono = "0.4"
 docopt = "1.1"
-cloud-storage = "0.6" # 0.7+ includes request 0.11, tokio 1.4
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ bytes = "1.0"
 cadence = "0.26"
 chrono = "0.4"
 docopt = "1.1"
-cloud-storage = "0.6" # 0.7+ includes request 0.11, tokio 1.4
+cloud-storage = { git = "https://github.com/mozilla-services/cloud-storage-rs", branch = "feat/create_with_params-0.6.2" } # 0.7+ includes request 0.11, tokio 1.4
```
pjenvey, line 29: this branch could use a separate review, I can set up a PR to merge it to another branch for that sake

src/server/img_storage.rs
```diff
@@ -121,7 +121,7 @@ impl Default for StoreImage {
 #[derive(Debug)]
 pub struct StoredImage {
     pub url: uri::Uri,
-    pub object: Object,
+    pub object: Option<Object>,
```
jrconlin, line 124: we can probably drop this. It's not really used.

jrconlin (approved):

jrconlin:
Once this lands, really tempted to redo #258 to drop the useless vendor and just be "use redis as a shared image metadata cache". It'll still be a draft we can keep in our collective back pocket, but i'll note the redis integration was fairly easy.

jrconlin (approved):

src/server/img_storage.rs
pjenvey: woops, almost forgot this

pjenvey: And then I actually did forget this (forgot to push it)

PR #264, opened 2021-08-17
fix: re-add location_test_header support

### LINKED ISSUE #257: Respect location_test_header w/ maxmind enabled
With maxmind enabled, we skip reading the test location header. When enabled the test location header should trump all other providers.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1203)

pjenvey:
## Description

via its own location provider, checked before maxmind

(draft for now in case it conflicts w/ #235)

## Testing

New tests pass

## Issue(s)

Closes #257


jrconlin (dismissed):
I do wonder if we should put this behind either a feature flag or a setting so that it's only enabled for load testing configurations. I don't think it's a huge threat, but someone could screw with things by adding their own `X-Test-Location` header to calls. (Maybe we flag these as "testing" and don't count them?)

pjenvey:
> I do wonder if we should put this behind either a feature flag or a setting so that it's only enabled for load testing configurations.

It is conditional behind the same `test_location_header` setting as it was before

pjenvey:
Moving out of draft as I don't think it conflicts much at all w/ #235

jrconlin (approved):
?? wonder why that didn't transfer?

pjenvey:
> ?? wonder why that didn't transfer?

Manually merged it for conflicts and added another commit on top of that

PR #271, opened 2021-08-19
feat: support dma-codes in location_test_header

### LINKED ISSUE #269: Support dma-code in location_test_header
Follow up to #205:

The location test header should allow testing clients to additionally specify a dma-code, ensuring we're able to better simulate a cache similar to the production environment's.

Then we'll need a new issue for contile-loadtest to utilize it, sending random dma-codes (or omitting dma codes) for US regions. Ideally they wouldn't be _too_ random -- there should only so many DMA codes used per individual US state.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1226)

pjenvey:
## Issue(s)

Closes #269

### DIFF
--- a/src/server/location.rs
+++ b/src/server/location.rs
@@ -58,6 +58,15 @@ impl Provider for TestHeaderProvider {
                 }
             }
 
+            if let Some(dma) = parts.next() {
+                let dma = dma.trim().parse().unwrap_or(0);
+                // Non-exact validation (there's only 210 DMA regions) but
+                // close enough for testing
+                if (500..=900).contains(&dma) {
+                    builder = builder.dma(dma);
+                }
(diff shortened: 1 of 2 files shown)


src/server/location.rs
```diff
@@ -58,6 +58,15 @@ impl Provider for TestHeaderProvider {
                 }
             }
 
+            if let Some(dma) = parts.next() {
```
jrconlin, line 61: Ah, I see what you're doing. I wonder if we might want to make this more a JSON blob so there's no threat of screwing up order or something. That could definitely be a different ticket, though.

jrconlin (approved):

PR #273, opened 2021-08-20
Add integration test for legacy image filter

hackebrot:
## Description

This pull request adds a new integration test to verify that Contile successfully filters tiles based on whether there's an image for the ad provider of a tile shipped with Firefox 90.

- I have added a new ad provider for testing named `DunBroch` which is only returned for in requests from `GB`
- Only `Example ORG` and `Example COM` have  legacy images (see `docker-compose.yml`)
- The first request in the new scenario comes from Firefox 91 and we expect tiles for `DunBroch` as well as `Example COM`
- The second request in the new scenario comes from Firefox 90 and we expect **no** tiles for `DunBroch`

See https://github.com/mozilla-services/contile-integration-tests/pull/64#pullrequestreview-732978460

## Testing

Verify that the new test `legacy_image_filter` runs on CI.

## Issue(s)

The feature under test was added in #247.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,7 +24,7 @@ services:
     environment:
       CONTILE_MAXMINDDB_LOC: /tmp/mmdb/GeoLite2-City-Test.mmdb
       CONTILE_ADM_ENDPOINT_URL: http://partner:5000/tilesp
-      CONTILE_ADM_QUERY_TILE_COUNT: 2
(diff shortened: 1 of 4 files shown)


ashrivastava-qa (approved):
LGTM!

ncloudioj (approved):
That's exactly what we want to cover, thanks!

jrconlin (approved):

hackebrot:
Thanks for the reviews, team! 🚀

PR #304, opened 2021-10-15
feat: add regex rule parser for advertiser hosts

### LINKED ISSUE #303: Add subdomain allowlist to Contile
Business requirement:

Apart from the main (top) domains, adM wants us to support multiple subdomains (e.g christmas.homedepot.com) for certain advertisers.

Contile currently already allows all the subdomains "*.foo.com" for "foo.com", we'd like to add some control on the subdomain allowlist instead of letting adM add arbitrary subdomains without Mozilla's approval.

Put another way, for any approved advertiser, in addition to its top domain "foo.com", we can also specify a list of allowed subdomains in Contile such as "bar.foo.com", "baz.foo.com", etc. Contile will not treat those landing URLs as invalid. 

Note:
For some advertisers that don't use subdomains but paths, for example, Etsy in Canada has the landing URL set as "www.etsy.com/ca", this subdomain allowlist scheme will not work.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1309)

jrconlin:
This will allow for pattern matches against `advertiser_hosts` (See
 #303). Regular expression patterns should be prefixed with an "r#"
e.g.
```json
    "advertiser_hosts": ["r#^(www\\.)([\\w\\.]+)?acme\\.biz/(ca|uk|us|ie)/"],
```

Traditional "host only" filters should still work
e.g
```json
   "advertiser_hosts": ["acme.biz", "acme.ca"],
```

Also, not a great idea to mix and match regex and traditional style.
For help generating valid regular expressions, you can use sites like
https://regexr.com/

Closes #303


jrconlin:
Yeah, not really going to complain. Regex is really powerful, a resource hog, and made out of pure explodium. I think we're probably NOT going to land this PR. I created it mostly as a draft to see what all would be involved in actually creating it, and realized I had accidentally built a patch. 

There's a working document that I'm using with our customer to create a simpler string based rule set that should solve their requirements without invoking the great satan that regex can be.

hackebrot:
Hi @jrconlin! 👋🏻 Would you say this new feature is sufficiently covered by unit tests? Do you think we can add at least one integration test that uses settings which rely on this feature?

PR #311, opened 2021-11-04
feat: Refine the host & pathing filter

ncloudioj:
After a second thought, I don't think letting the content partner send whatever URLs to us is a good idea, so I took a crack at refining the existing pathing filter with the following changes:
* Rename the `advertiser_hosts` to `advertiser_urls`
* `advertiser_urls` now takes a list of hosts and path filters (optional). If no path filter is specified, it will only check the host as the current behavior
* Each path filter could be either an "exact-matching" filter or a "prefix-matching" one
* The "value" field for "prefix-matching" filters must contain the trailing "/" in the configuration, otherwise, Contile will panic

The detail of the filtering is documented [here](https://github.com/mozilla-services/contile/pull/311/files#diff-34387244eabf61c959f3a7e535ca04a551c779b9ee645d2e93303c0020f5a881).

Note: this is a breaking change, we'd need to update the Contile configuration file if we want to deploy this.

@hackebrot could you also take a look at the change in the integration test? I am not sure if I covered everything there. Thanks!

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -531,13 +531,13 @@ mod tests {
                 {
                     "host": "acme.biz",
                     "paths": [
-                        { "value": "/ca", "matching": "prefix" }
+                        { "value": "/ca/", "matching": "prefix" }
                     ]
                 },
                 {
                     "host": "www.acme.co",
                     "paths": [
-                        { "value": "/foo.bar", "matching": "prefix" }
+                        { "value": "/foo.bar/", "matching": "prefix" }
                     ]
 
                 },
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -115,25 +115,35 @@ impl AdmFilter {
         }
 
         // do a quick string comparison between the supplied host and the provided filter.
-        let mut host = format!(
-            "{}{}",
-            parsed
-                .host()
-                .ok_or_else(|| HandlerErrorKind::MissingHost(species, url.to_string()))?,
-            parsed.path()
-        );
-        if !host.ends_with('/') {
-            host.push('/');
+        let host = parsed
+            .host()
+            .ok_or_else(|| HandlerErrorKind::MissingHost(species, url.to_string()))?
+            .to_string();
+        let mut path = Cow::from(parsed.path());
+        if !path.ends_with('/') {
+            path.to_mut().push('/');
         }
-        for filter in &filter.advertiser_hosts {
-            let mut filter = Cow::from(filter);
-            if !filter.ends_with('/') {
-                filter.to_mut().push('/');
-            };
-            if host.starts_with(filter.as_ref()) {
-                return Ok(());
+
+        for filter in &filter.advertiser_urls {
+            if !host.eq(&filter.host) {
+                continue;
             }
+
+            if let Some(ref paths) = filter.paths {
+                for rule in paths {
+                    match rule.matching.as_str() {
+                        // Note that the orignal path is used for exact matching
+                        "exact" if rule.value == parsed.path() => return Ok(()),
+                        "prefix" if path.starts_with(&rule.value) => return Ok(()),
+                        _ => continue,
+                    }
+                }
+            } else {
+                // Host matches without any path filters, matching succeeds.
+                return Ok(());
+            };
         }
+
         Err(HandlerErrorKind::InvalidHost(species, url.to_string()).into())
     }
 
@@ -294,7 +304,7 @@ impl AdmFilter {
                     return None;
                 }
 
-                let adv_filter = if filter.advertiser_hosts.is_empty() {
+                let adv_filter = if filter.advertiser_urls.is_empty() {
                     default
                 } else {
                     filter
@@ -433,7 +443,27 @@ mod tests {
     #[test]
     fn check_advertiser() {
         let s = r#"{
-            "advertiser_hosts": ["acme.biz/ca", "black_friday.acme.biz/ca"],
+            "advertiser_urls": [
+                {
+                    "host": "acme.biz",
+                    "paths": [
+                        { "value": "/ca/", "matching": "prefix" }
+                    ]
+                },
+                {
+                    "host": "black_friday.acme.biz",
+                    "paths": [
+                        { "value": "/ca/", "matching": "prefix" }
+                    ]
+
+                },
+                {
+                    "host": "acme.biz",
+                    "paths": [
+                        { "value": "/usa", "matching": "exact" }
+                    ]
+                }
+            ],
             "position": 0
         }"#;
         let settings: AdmAdvertiserFilterSettings = serde_json::from_str(s).unwrap();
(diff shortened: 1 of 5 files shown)


src/adm/filter.rs
```diff
@@ -477,15 +507,47 @@ mod tests {
...
+            .is_err());
+
         // "Traditional host. "
         let s = r#"{
-            "advertiser_hosts": ["acme.biz/ca", "www.acme.co/foo.bar"],
+            "advertiser_urls": [
+                {
+                    "host": "acme.biz",
+                    "paths": [
+                        { "value": "/ca", "matching": "prefix" }
```
jrconlin, line 534: shouldn't these panic? https://github.com/mozilla-services/contile/pull/311/files#diff-34387244eabf61c959f3a7e535ca04a551c779b9ee645d2e93303c0020f5a881R50

src/adm/settings.rs
```diff
@@ -17,6 +17,52 @@ use crate::{
...
+///   then the URL is accepted by this filter.
+/// * If the `paths` filter list is present, then proceed with path filtering.
+///   There are two "matching" strategies:
+///   * "exact" for exact path matching, which compares the `path` character-by-character
+///     with the `value` filed of this path filter.
+///   * "prefix" for prefix path matching, which checks if the `value` is a
+///     prefix of the `path`. Note that we always make sure `path` and `value`
+///     are compared with the trailing '/' to avoid the accidental
+///     matches. In particular, Contile will panic when it detects that a prefix
+///     filter doesn't have the trailing '/' in the `value`.
```
jrconlin, line 51: I don't believe that this line is true. Your tests include several "prefix" paths that do not end with slashes and there is no panic.

ncloudioj: Nice catch, that was inaccurate. It will be only enforced when loading the filters from the configuration file. I will update it.

src/adm/filter.rs
ncloudioj: Those "values" are wrong, they should have trailing slashes.

jrconlin (approved):

jrconlin:
I think there's still some potential problems we could hit (e.g. combining `exact` and `prefix`, using empty `prefix`, etc.) but we can handle those in a separate PR.

ncloudioj:
@jrconlin Thanks for the review!

> I think there's still some potential problems we could hit (e.g. combining `exact` and `prefix`, using empty `prefix`, etc.) but we can handle those in a separate PR.

Agreed, there is currently no validation on those filters except the trailing slash check. Will do that in a follow-up.

@hackebrot let me know if those changes in the integration config file make sense to you? I will merge once you give 👍

hackebrot (approved):
Hi @ncloudioj! 👋🏻 

Based on my understand of the [documentation you've linked](https://github.com/mozilla-services/contile/pull/311/files#diff-34387244eabf61c959f3a7e535ca04a551c779b9ee645d2e93303c0020f5a881), the changes to `integration-tests/volumes/contile/adm_settings.json` look correct to me!

I do wonder if it makes sense to also update `click_hosts` and `impression_hosts` to match the updated structure for `advertister_hosts` for the sake of consistency. But that's not a blocker for merging your pull request.

This is a great opportunity for us to increase our meaningful test coverage and verify the different path matching mechanisms in the automated integration tests. I can create a follow-up pull request for this. 🤖

hackebrot:
@ncloudioj It seems contile on this branch doesn't panic when a path value does **not** have a trailing `/` in the settings JSON file. 🤔 

I see the unit test passed on CI:

```text
test adm::settings::tests::test_invalid_path_filters - should panic ... ok
```

And the unit tests in this pull request pass for me locally as well:

```text
cargo test adm::settings
```

```text
running 4 tests
test adm::settings::tests::test_invalid_path_filters ... ok
test adm::settings::tests::test_obsolete_settings ... ok
test adm::settings::tests::test_lower_ignore ... ok
test adm::settings::tests::all_include_regions ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 36 filtered out; finished in 0.00s
```

But the following diff does not seem to cause contile to panic. All integration tests pass and I cannot find `PathFilter` in the log messages from contile even though that's in the panic message:

```diff
     "DunBroch": {
         "advertiser_urls": [
             {
-                "host": "www.dunbroch.co.uk"
+                "host": "www.dunbroch.co.uk",
+                "paths": [
+                    {
+                        "value": "/gb_desktop",
+                        "matching": "prefix"
+                    }
+                ]
             }
         ],
         "click_hosts": [
```

However the following does produce 1 integration test failure (`legacy_image_filter`):

```diff
     "DunBroch": {
         "advertiser_urls": [
             {
-                "host": "www.dunbroch.co.uk"
+                "host": "www.dunbroch.co.uk",
+                "paths": [
+                    {
+                        "value": "/gb_desktop_NOPE",
+                        "matching": "prefix"
+                    }
+                ]
             }
         ],
         "click_hosts": [
```

Because contile filters out `https://www.dunbroch.co.uk/gb_desktop_macos` and returns `https://www.example.com/gb_desktop_macos` instead. But it also doesn't contile to panic and contile does return tiles.

ncloudioj:
> I do wonder if it makes sense to also update `click_hosts` and `impression_hosts` to match the updated structure for `advertister_hosts` for the sake of consistency. But that's not a blocker for merging your pull request.

Yes, we could do to be consistent. but I don't think `click_hosts` and `impression_hosts` are as critical as `advertiser_url` for two reasons:
* `advertiser_url` is user-facing, whereas the other two are only used by our data pipeline. So the level of risk is much lower.
* Secondly, it's not uncommon for the partner to update the `click_hosts` and `impression_hosts` on their side, they're much more volatile than advertiser's landing URLs. It'll be much harder for us to keep in sync, that's why both Contile and data pipeline only check the host other than the full path. That being said, if we make the same change to `click_hosts` and `impression_hosts`, chances are we will stick to the host matching like the current one.

ncloudioj:
> But the following diff does not seem to cause contile to panic. All integration tests pass and I cannot find `PathFilter` in the log messages from contile even though that's in the panic message:

@hackebrot Did you try this as an unit test in `filter.rs`? 

As noted in the [doc](https://github.com/mozilla-services/contile/pull/311/files#diff-34387244eabf61c959f3a7e535ca04a551c779b9ee645d2e93303c0020f5a881), the validation of the path `value` is only performed on the setting parsing phase, as to my understanding that is the only place Contile reads the filter configuration in production. After that, it won't check the validity on `value` any more.

hackebrot:
> 
> @hackebrot Did you try this as a unit test in `filter.rs`?
> 

Not yet, but the the behaviour can be observed in #313.

FWIW the following changes for `check_advertiser` in `src/adm/filter.rs` do **not** cause the unit tests to fail:

```diff
                 {
                     "host": "black_friday.acme.biz",
                     "paths": [
-                        { "value": "/ca/", "matching": "prefix" }
+                        { "value": "/ca", "matching": "prefix" }
                     ]
 
                 },
```

jrconlin:
> > @hackebrot Did you try this as a unit test in `filter.rs`?
>
> FWIW the following changes for `check_advertiser` in `src/adm/filter.rs` do **not** cause the unit tests to fail:
> 

I think there might be a bit of confusion about this. As I understand (deferring to Nan), the `panic!` happens in [settings](https://github.com/mozilla-services/contile/pull/311/files#diff-34387244eabf61c959f3a7e535ca04a551c779b9ee645d2e93303c0020f5a881R200) when the config file is read in. After that, a panic should not happen, so (out of an abundance of precaution) [filter adds a terminating `/`](https://github.com/mozilla-services/contile/pull/311/files#diff-3df4b5b49a5986eab0838fe5d60597b514e75784fdda4b18bb2d4393459d7846R123) to the path pattern. 

That means that if you call `filter.check_advertiser` with a settings that you've hand built in a test that may not contain a terminal slash, `check_advertiser` would "fix" the error anyway. I believe that the original PR had this in place and I noted the discrepancy. Nan fixed it in a latter commit.

ncloudioj:
> That means that if you call `filter.check_advertiser` with a settings that you've hand built in a test that may not contain a terminal slash, `check_advertiser` would "fix" the error anyway. I believe that the original PR had this in place and I noted the discrepancy. Nan fixed it in a latter commit.

Exactly, the intent here is that we wanted to make sure the prefix values are always provided with the trailing "/" in the settings, if not, Contile should refuse to boot by panicking.

PR #313, opened 2021-11-05
Integration test coverage for path filtering

hackebrot:
## Description

This pull request is a follow-up for #311 and uses that branch as a base. I will rebase this once #311 is merged.

Please see the test scenario descriptions in `integration-tests/volumes/client/scenarios.yml` and let me know if you have any questions about the scenarios.

Note that I've observed unexpected behavior when working this, which I described in https://github.com/mozilla-services/contile/pull/311#issuecomment-961752700. 🚧 

## Testing

Verify that the following 2 new integration tests run and pass on CI:

* `advertiser_url_path_filter_prefix`
* `advertiser_url_path_filter_exact`

## Issue(s)

See #311

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -214,13 +214,30 @@ jobs:
             -f integration-tests/docker-compose.yml \
             up --abort-on-container-exit --force-recreate
       - run:
-         name: Run additional integration tests
+         name: 'Run "204" integration tests'
          command: |
            docker-compose --version
            docker-compose \
             -f integration-tests/docker-compose.yml \
             -f integration-tests/docker-compose.204.yml \
             up --abort-on-container-exit --force-recreate
+      - run:
+         name: 'Run "init_error" integration tests'
+         command: |
+           docker-compose --version
+           set +e # We need this so that the run doesn't exit after docker-compose
+           docker-compose \
+            -f integration-tests/docker-compose.yml \
+            -f integration-tests/docker-compose.init_error.yml \
+            up --abort-on-container-exit --exit-code contile --force-recreate
+           contile_exit_code=$?
+           if [ "${contile_exit_code}" -eq 0 ]; then
+            echo "Expected non-zero exit_code from Contile service"
+            exit 1
+           else
+            echo "Contile service exit_code: ${contile_exit_code}"
+            exit 0
+           fi
 
 workflows:
   version: 2
--- a/integration-tests/docker-compose.init_error.yml
+++ b/integration-tests/docker-compose.init_error.yml
@@ -1,0 +1,9 @@
+version: "3"
+
+services:
+  contile:
+    environment:
+      CONTILE_ADM_SETTINGS: /tmp/contile/adm_settings_init_error.json
+  client:
+    environment:
+      SCENARIOS_FILE: /tmp/client/scenarios_init_error.yml
--- a/integration-tests/volumes/client/scenarios.yml
+++ b/integration-tests/volumes/client/scenarios.yml
@@ -316,3 +316,88 @@ scenarios:
           status_code: 200
           content:
             tiles: []
+
+  - name: advertiser_url_path_filter_prefix
+    description: >
+      Test that Contile successfully applies advertiser_url path filters with
+      prefix matching. The request is expected to not result in any DunBroch
+      tiles because the settings do not allow URL paths with a
+      `gb_desktop_windows` prefix for DunBroch tiles.
+    steps:
+      - request:
+          method: GET
+          path: '/v1/tiles'
+          headers:
+            # Contile maps the User-Agent Header value to os-family and form-factor parameters
+            # The following value will result in os-family: windows and form-factor: desktop
+            - name: User-Agent
+              value: 'Mozilla/5.0 (Windows NT 10.0; rv:10.0) Gecko/20100101 Firefox/91.0'
+            # Contile looks up the IP address from this header value and maps it to proxy information.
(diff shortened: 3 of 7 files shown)


hackebrot:
I added a new integration test that uses a `adm_settings.json` file with a prefix path filter with a value without a trailing `/` which I expect to cause Contile to panic. My understanding of how panics are handled in Contile or its underlying web framework may be inaccurate.

Do you expect tiles from the advertiser hosts with invalid path filters to be ignored? I'm seeing a `204 No Content`.

hackebrot:
It seems that the `panic!` in Contile when loading the settings JSON does not cause the entire `actix-web` server to terminate, but only the worker that encounters that panic. So invalid settings will not cause the server to terminate and clients can still send requests to the server.

jrconlin:
Ah! Good catch!

The panic would get recorded by Sentry, and the code enforces a terminating slash. 
@ncloudioj, I'm tempted to have settings record it as either a new `ConfigurationError`, or record it as a warning and have the settings module correct the path when it encounters it, instead of panicking.

ncloudioj:
> Ah! Good catch!
> 
> The panic would get recorded by Sentry, and the code enforces a terminating slash.

That's quite surprising to me. I'd guess all the Actix servers (i.e. workers) would load the same settings, shouldn't they all panic? How did it serve the request and return 200 🤔 

> @ncloudioj, I'm tempted to have settings record it as either a new `ConfigurationError`, or record it as a warning and have the settings module correct the path when it encounters it, instead of panicking.

Yep, I think we can do that although it'll be preferred to fail soon and loud.

hackebrot:
> That's quite surprising to me. I'd guess all the Actix servers (i.e. workers) would load the same settings, shouldn't they all panic? How did it serve the request and return 200 🤔

Based on the results in the integration tests, I am wondering if only workers that happen to process tiles from partner API for an advertiser for which there is an invalid path filter in the settings panic and all other workers are not impacted. Could that be the case? Does settings loading happen lazily?

pjenvey:
> I added a new integration test that uses a `adm_settings.json` file with a prefix path filter with a value without a trailing `/` which I expect to cause Contile to panic. My understanding of how panics are handled in Contile or its underlying web framework may be inaccurate.

I'm not seeing any panic running the `more-path-filtering` branch w/ `CONTILE_ADM_SETTINGS=adm_settings_panic.json` -- seems like it's not triggering it.

Indeed rust's panic will only affect the current thread (potentially only one actix-web worker while the others continue). However contile's initialization of settings occur within `server::Server::with_settings` outside of actix-web's workers. An error there should panic the main (and only thread at that time) thread, aborting the entire program.

integration-tests/volumes/client/scenarios_panic.yml
```diff
@@ -0,0 +1,23 @@
+scenarios:
+  - name: advertiser_url_prefix_path_filter_missing_slash
```
jrconlin, line 2: I don't think this tests what we think it does.

If the settings are invalid, contile should fail to launch, so any query would fail. 

I believe we test that with `docker-compose.panic.yml`.

jrconlin (changes requested):

integration-tests/volumes/client/scenarios_panic.yml
hackebrot: Hi @jrconlin! 👋🏻 

I wrote these tests before we concluded that the main thread does not panic on Slack meaning that it wasn't clear what behavior we expect from Contile if the settings are invalid with #311. This tests does demonstrate that Contile didn't not panic when the settings are passed in as a file path.

I will update this now.

ncloudioj:
FWIW, #314 has removed this panic, instead, it simply returns an error when detecting malformed path filters (e.g. prefix values without the trailing "/" or exact values without the leading "/").

pjenvey:
> FWIW, #314 has removed this panic, instead, it simply returns an error when detecting malformed path filters (e.g. prefix values without the trailing "/" or exact values without the leading "/").

It does still "panic" though -- for the sake of this PR the behavior is the same (it was changed to return Errors but such initialization errors are propagated out to contile's main function, ultimately causing it to abort w/ a non zero return code before the server is launched).

hackebrot:
I rebased this branch to reorder, fixup and reword some of the commits. I also renamed what we originally called `panic` integration tests to `init_error` to make it clear that we're testing for an unrecoverable error during Contile's initialization as @pjenvey correctly pointed out. In this case the init error is because of invalid contents in the settings file.

ncloudioj (approved):
LGTM, thanks!

.circleci/config.yml
```diff
@@ -214,13 +214,30 @@ jobs:
...
            docker-compose \
             -f integration-tests/docker-compose.yml \
             -f integration-tests/docker-compose.204.yml \
             up --abort-on-container-exit --force-recreate
+      - run:
+         name: 'Run "init_error" integration tests'
+         command: |
+           docker-compose --version
+           set +e # We need this so that the run doesn't exit after docker-compose
+           docker-compose \
```
jrconlin, line 229: Not needed now, but you can use `timeout --kill-after=DURATION --signal=SIGNAL ...` to automatically kill a command with SIGNAL after DURATION seconds.

hackebrot: Okay. Here with `--abort-on-container-exit` docker compose will shut down all services as soon as the `contile` service exits or if that doesn't happen after the `client` exits because the test ran.

ashrivastava-qa (approved):
LGTM 👍

PR #314, opened 2021-11-08
bug: ensure settings from file are filtered like env settings

jrconlin:
There is a bug where ADM settings read from a file would return early and NOT receive the additional checks. 

This extends https://github.com/mozilla-services/contile/pull/311

### DIFF
--- a/src/adm/settings.rs
+++ b/src/adm/settings.rs
@@ -28,8 +28,8 @@ pub(crate) const DEFAULT: &str = "DEFAULT";
 ///         "host": "foo.com",
 ///         "paths": [
 ///             { "value": "/", "matching": "exact" },
-///             { "value": "/bar", "matching": "prefix" },
-///             { "value": "/baz/spam", "matching": "prefix" },
+///             { "value": "/bar/", "matching": "prefix" },
+///             { "value": "/baz/spam/", "matching": "prefix" },
 ///         ]
 ///     }
 /// ```
@@ -47,7 +47,8 @@ pub(crate) const DEFAULT: &str = "DEFAULT";
 ///   * "prefix" for prefix path matching, which checks if the `value` is a
 ///     prefix of the `path`. Note that we always make sure `path` and `value`
 ///     are compared with the trailing '/' to avoid the accidental
-///     matches.
+///     matches. In particular, Contile will panic when it detects that a prefix
+///     filter doesn't have the trailing '/' in the `value`.
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct AdvertiserUrlFilter {
     pub(crate) host: String,
--- a/src/adm/settings.rs
+++ b/src/adm/settings.rs
@@ -1,9 +1,13 @@
 use std::{
     collections::{HashMap, HashSet},
+    convert::TryFrom,
     fmt::Debug,
     fs::File,
+    io::Read,
     path::Path,
 };
+
+use config::ConfigError;
 
 use serde::{ser::SerializeSeq, Deserialize, Deserializer, Serialize, Serializer};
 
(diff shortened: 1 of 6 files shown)


src/adm/settings.rs
```diff
@@ -168,26 +177,46 @@ impl From<&mut Settings> for AdmSettings {
...
+                if let Some(ref paths) = filter.paths {
+                    return paths
+                        .iter()
+                        .any(|path| path.matching == "prefix" && !path.value.ends_with('/'));
+                }
+                false
+            }) {
+                return Err(ConfigError::Message(format!("Advertiser {:?} advertiser_urls contain invalid prefix PathFilter (missing trailing '/')", adv)));
             }
             if filter_setting.advertiser_urls.iter().any(|filter| {
```
ncloudioj, line 221: Did you want to remove this block now as it already returns an error above?

src/adm/settings.rs
```diff
@@ -168,26 +177,46 @@ impl From<&mut Settings> for AdmSettings {
...
+                return Err(ConfigError::Message(format!(
+                    "Advertiser {:?} include_regions must be uppercase",
+                    adv
+                )));
+            }
+            if filter_setting.advertiser_urls.iter().any(|filter| {
+                if let Some(ref paths) = filter.paths {
+                    return paths
+                        .iter()
+                        .any(|path| path.matching == "prefix" && !path.value.ends_with('/'));
```
ncloudioj, line 215: While you're refactoring the code here, could you also add a check for "exact" values? Specifically, valid "exact" values should start with '/'.

ncloudioj: To make things easier, I think we can just use a single error message to capture both, such as `Advertiser {:?} advertiser_urls contain invalid path values`

ncloudioj (approved):
Looks great, r+ with review comments.

pjenvey (changes requested):

src/adm/settings.rs
```diff
@@ -168,26 +177,46 @@ impl From<&mut Settings> for AdmSettings {
...
                 .iter()
                 .any(|region| region != &region.to_uppercase())
             {
-                panic!("Advertiser {:?} include_regions must be uppercase", adv);
+                return Err(ConfigError::Message(format!(
+                    "Advertiser {:?} include_regions must be uppercase",
+                    adv
+                )));
+            }
+            if filter_setting.advertiser_urls.iter().any(|filter| {
```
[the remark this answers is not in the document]
jrconlin: removed duplication.

src/adm/settings.rs
```diff
@@ -59,12 +63,44 @@ pub struct AdvertiserUrlFilter {
     pub(crate) paths: Option<Vec<PathFilter>>,
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize)]
```
pjenvey, line 66: nit: could add `Copy` as well

src/adm/settings.rs
```diff
@@ -59,12 +63,44 @@ pub struct AdvertiserUrlFilter {
...
+            "exact" => Ok(Self::Exact),
+            _ => Err(ConfigError::Message(format!(
+                "Invalid Path Filter Type {}",
+                string
+            ))),
+        }
+    }
+}
+
+impl From<PathMatching> for String {
```
pjenvey, line 88: total nit because I don't even think this is used but this impl could be `for &'static str` to avoid the allocation.

pjenvey (dismissed):
r+ w/ a couple nits on the new enum

src/adm/settings.rs
jrconlin: Yeah, I do these out of habit because I inevitably need them later.
Thanks for the tip, though. I'll do that.

pjenvey (approved):

PR #331, opened 2021-12-07
feat: Read ADM settings data from a Google Storage bucket.

### LINKED ISSUE #324: Optionally read ADM Setting from a Cloud storage bucket
In support of the Contile settings manager, we should read the ADM settings data from a Google Storage Bucket file. 

This data should be in JSON format (to standardize with other input methods) and may optionally re-fetch data from the bucket on a period basis. 

Invalid or incorrectly formatted data should not replace existing settings.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1389)
┆epic: Contile Settings Manager

jrconlin:
Note: This only reads the brand filter data, other ADM data will require
redeployment.

Closes #324

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -55,7 +55,8 @@ pub struct AdmFilter {
     /// for pre 91 tile support.
     pub legacy_list: HashSet<String>,
     pub source: String,
-    pub last_updated: Option<std::time::SystemTime>,
+    pub source_url: Option<url::Url>,
+    pub last_updated: Option<chrono::DateTime<chrono::Utc>>,
     pub refresh_rate: std::time::Duration,
 }
 
@@ -103,7 +104,10 @@ fn check_url(url: Url, species: &'static str, filter: &[Vec<String>]) -> Handler
 impl AdmFilter {
     /// convenience function to determine if settings are cloud ready.
     pub fn is_cloud(&self) -> bool {
-        self.source.starts_with("gs://")
+        if let Some(source) = &self.source_url {
+            return source.scheme() == "gs";
+        }
+        false
     }
 
     /// Report the error directly to sentry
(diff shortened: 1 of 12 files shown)


src/adm/filter.rs
```diff
@@ -107,6 +115,104 @@ impl AdmFilter {
         l_sentry::report(&merged_tags, sentry::event_from_error(error));
     }
 
+    /// check to see if the bucket has been modified since the last time we updated.
+    pub async fn requires_update(&self) -> HandlerResult<bool> {
+        // don't update non-bucket versions (for now)
+        if !self.is_cloud() {
+            return Ok(false);
+        }
+        let bucket: url::Url = match self.source.parse() {
```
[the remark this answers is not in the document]
jrconlin: Ah, good call. I had done it as a String mostly to get around the `Default`, and because the source could be a file path and not a URL. I"ll add the extra logic to handle that cleanly.

src/adm/filter.rs
```diff
@@ -107,6 +115,104 @@ impl AdmFilter {
...
+            return Ok(chrono::DateTime::<chrono::Utc>::from_utc(
+                chrono::NaiveDateTime::from_timestamp(
+                    updated
+                        .duration_since(std::time::UNIX_EPOCH)
+                        .unwrap_or_else(|_| std::time::Duration::from_secs(0))
+                        .as_secs() as i64,
+                    0,
+                ),
+                chrono::Utc,
+            ) <= obj.updated);
```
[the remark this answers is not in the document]
jrconlin: Partly because I was hitting issues with the result being in DateTime and the UTC being in SystemTime, and just being frustrated about it all. Good call, I'll fix it.

src/adm/filter.rs
```diff
@@ -107,6 +119,84 @@ impl AdmFilter {
...
+        if let Some(bucket) = &self.source_url {
+            let adm_settings = AdmSettings::from_settings_bucket(bucket)
+                .await
+                .map_err(|e| {
+                    HandlerError::internal(&format!(
+                        "Invalid bucket data in {:?}: {:?}",
+                        self.source, e
+                    ))
+                })?;
+            for (adv, setting) in adm_settings.advertisers {
```
pjenvey, line 168: This should also handle an advertiser entry being removed upstream

src/adm/settings.rs
```diff
@@ -174,7 +173,104 @@ where
...
+        settings_bucket: &url::Url,
+    ) -> Result<AdmSettings, ConfigError> {
+        let settings_str = settings_bucket.as_str();
+        if settings_bucket.scheme() != "gs" {
+            return Err(ConfigError::Message(format!(
+                "Improper bucket URL: {:?}",
+                settings_str
+            )));
+        }
+        let bucket_name = match settings_bucket.host() {
```
pjenvey, line 252: Another `ok_or_else` usage here

src/adm/settings.rs
```diff
@@ -316,6 +400,13 @@ impl From<&mut Settings> for HandlerResult<AdmFilter> {
             ignore_list,
             all_include_regions,
             legacy_list,
+            last_updated: match &source.starts_with("gs://") {
+                true => Some(chrono::Utc::now()),
+                false => None,
+            },
```
pjenvey, line 406: They added this in Rust 1.50!
```suggestion
            last_updated: source.starts_with("gs://").then(|| chrono::Utc::now()),
```

src/adm/tiles.rs
```diff
@@ -211,13 +211,22 @@ pub async fn get_tiles(
         metrics.incr_with_tags("filter.adm.empty_response", Some(tags));
     }
 
+    // update the filters if needed.
+    if settings.adm_live_update && state.filter.read().unwrap().requires_update().await? {
```
pjenvey, line 215: This should be needed generally with the background updater.

I think the one time it might be needed is during startup if the background updater hasn't completed reading the bucket yet. Instead Could `Server::with_settings` call `AdmFilter::update`, blocking on starting the server until it's finished?

src/server/mod.rs
```diff
@@ -29,19 +30,33 @@ const REQWEST_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARG
...
 
+impl Clone for ServerState {
+    fn clone(&self) -> Self {
+        Self {
+            metrics: self.metrics.clone(),
+            adm_endpoint_url: self.adm_endpoint_url.clone(),
+            reqwest_client: self.reqwest_client.clone(),
+            tiles_cache: self.tiles_cache.clone(),
+            settings: self.settings.clone(),
+            filter: RwLock::new((*self.filter.read().unwrap()).clone()),
```
pjenvey, line 53: We don't want to clone the `AdmFilter` here, ending up with separate copies because the background updater only updates the original copy. I think we can wrap the `RwLock` in an `Arc` instead.

pjenvey (changes requested):

src/adm/tiles.rs
jrconlin: I was having problems with actix calling async calls from inside sync calls which lead to this. Basically, things compile, but then the app fails because handler is already in use. I tried several different solutions to get around that, but nothing worked.

src/adm/filter.rs
jrconlin: added `delete` flag to ADMFilter settings.

src/adm/filter.rs
```diff
@@ -107,6 +119,90 @@ impl AdmFilter {
...
+                .to_string();
+
+            let obj = cloud_storage::Object::read(&host, bucket.path())
+                .await
+                .map_err(|e| {
+                    HandlerError::internal(&format!(
+                        "Could not read bucket {:?}, {:?}",
+                        self.source, e
+                    ))
+                })?;
```
pjenvey, line 143: Also: `HandlerError`'s have a `From` for CloudStorage errors now:
```suggestion
                .await?;
```

src/adm/filter.rs
```diff
@@ -107,6 +119,90 @@ impl AdmFilter {
...
+        }
+        if let Some(bucket) = &self.source_url {
+            let host = bucket
+                .host()
+                .ok_or_else(|| {
+                    HandlerError::internal(&format!("Missing bucket Host {:?}", self.source))
+                })?
+                .to_string();
+
+            let obj = cloud_storage::Object::read(&host, bucket.path())
```
pjenvey, line 136: Testing this w/ e.g. `CONTILE_ADM_SETTINGS=gs://pjenvey-test/adm_settings.json` --

this `read` and the `Object::download` call in settings both fail w/ 

```
GoogleError { domain: "global", reason: NotFound, message: "No such object: pjenvey-test//adm_settings.json"`
```

but they work with the leading slash removed from `bucket.path()`

src/adm/tiles.rs
pjenvey: Most of the startup is within the async runtime so this looks doable, here's what I mean: https://github.com/mozilla-services/contile/compare/feat/324-bucket...feat/324-bucket-block-init

Though that branch will crash contile at startup if the bucket reading fails for whatever reason. I'm not sure if that's worse than ignoring it and serving empty tiles, though

src/adm/filter.rs
```diff
@@ -107,6 +119,90 @@ impl AdmFilter {
...
+        }
+        Ok(())
+    }
+
+    // background settings updater.
+    pub async fn spawn_updater(&self) {
+        if !self.is_cloud() {
+            return;
+        }
+        let mut s = self.clone();
```
pjenvey, line 189: This `clone`'s similar to the other one we removed in `server/mod.rs` -- we don't want it as it means the background updater is updating a separate copy of the `AdmFilter` that isn't otherwise used.

The background updater should switch to operating on I think that `Arc<RwLock<AdmFilter>>` as well.

src/adm/filter.rs
jrconlin: Huh, that worked for me. Will investigate.

src/adm/tiles.rs
jrconlin: Yeah, I was trying not to do that, partly because the code got ugly kind of fast. Although I might have a fix for that...

pjenvey (approved):

PR #337, opened 2021-12-14
feat: Make the partner timeouts "softer" during initial cache loading.

### LINKED ISSUE #336: Do not return 500 error when filling cache
When we're rolling out a new deploy, the caches need to be filled. This can result in a spike of 500 errors to the client, which can cause a mild panic attack with Ops. 

We should consider options to either return a softer error type (206?), an empty result, trigger a lag so that values are eventually filled and returned, or some other kinder behavior.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1430)

jrconlin:
Closes #336

### DIFF
--- a/src/adm/tiles.rs
+++ b/src/adm/tiles.rs
@@ -246,9 +246,13 @@ pub async fn get_tiles(
         .tiles
         .into_iter()
         .filter_map(|tile| {
-            state
-                .filter
-                .filter_and_process(tile, location, &device_info, tags, metrics)
+            state.filter.read().unwrap().filter_and_process(
+                tile,
+                location,
+                &device_info,
+                tags,
+                metrics,
+            )
         })
         .take(settings.adm_max_tiles as usize)
         .collect();
--- a/src/adm/tiles.rs
+++ b/src/adm/tiles.rs
@@ -1,4 +1,10 @@
-use std::{fmt::Debug, fs::File, io::BufReader, path::Path, time::Duration};
+use std::{
+    fmt::Debug,
+    fs::File,
+    io::BufReader,
+    path::Path,
+    time::{Duration, Instant},
+};
 
 use actix_http::http::header::{HeaderMap, HeaderValue};
 use actix_web_location::Location;
@@ -193,8 +199,24 @@ pub async fn get_tiles(
             .send()
             .await
             .map_err(|e| {
+                // If we're just starting up, we're probably swamping the partner servers as
+                // we fill the queue. Instead of returning a normal 500 error, let's
+                // return something softer to keep our SRE's blood pressure lower.
+                //
+                // We still want to track this as a server error later.
+                //
+                // TODO: Remove this after the shared cache is implemented.
+                let mut err: HandlerError = if e.is_timeout()
+                    && Instant::now()
+                        .checked_duration_since(state.start_up)
+                        .unwrap_or_else(|| Duration::from_secs(0))
+                        <= Duration::from_secs(state.settings.adm_timeout)
+                {
+                    HandlerErrorKind::AdmLoadError().into()
+                } else {
+                    HandlerErrorKind::AdmServerError().into()
+                };
                 // ADM servers are down, or improperly configured
-                let mut err: HandlerError = HandlerErrorKind::AdmServerError().into();
                 err.tags.add_extra("error", &e.to_string());
                 err
             })?
--- a/src/adm/tiles.rs
+++ b/src/adm/tiles.rs
@@ -193,7 +193,8 @@ pub async fn get_tiles(
             AdmTileResponse::fake_response(&state.settings, test_response)?
         }
         crate::settings::TestModes::TestTimeout => {
-            return Err(HandlerErrorKind::AdmLoadError().into())
+            trace!("### Timeout!");
+            return Err(HandlerErrorKind::AdmLoadError().into());
         }
         _ => {
             state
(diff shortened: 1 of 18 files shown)


src/error.rs
```diff
@@ -94,6 +98,7 @@ impl HandlerErrorKind {
         match self {
             HandlerErrorKind::Validation(_) => StatusCode::BAD_REQUEST,
             HandlerErrorKind::AdmServerError() => StatusCode::SERVICE_UNAVAILABLE,
+            HandlerErrorKind::AdmLoadError() => StatusCode::PARTIAL_CONTENT,
```
ncloudioj, line 101: The Contile integration code in Firefox currently only handles HTTP 200 and 204. Given a 206, it will still try parsing the JSON from the response, which basically would share the same code path as HTTP 500.

Does it make sense to return HTTP 204 here instead?

ncloudioj (dismissed):
r+ w/ a comment.

src/error.rs
hackebrot: Do you mind adding integration tests for this, @jrconlin? 🤖

jrconlin: Returning a 204 is fine. I thought we had suggested 206 initially, but it doesn't matter from the server.

I'm not sure of the best way to trigger the test, mostly because this relies on an external failure state. I can see if I can construct a viable test around this, but it might be too artificial to be useful.

pjenvey: The integration test suite supports adding a delay (like a `time sleep()`) to responses, that could trigger the `Populating` state and possibly a subsequent step in the same scenario could then trigger this case?

pjenvey: My mistake, the second step is blocked on the first step finishing, so that wouldn't work

jrconlin:
With https://github.com/mozilla-services/contile/pull/337/commits/f173d2517c5a095cab0df4ae341d5e1d86a8f91e, I expanded the `CONTILE_TEST_MODE` flag to indicate states to run the server. I'm not sure how to best specify that using the existing integration tests.

I updated the old, crappy one to illustrate how to check for it.

pjenvey (changes requested):
I'm really sorry, I only glanced at this PR last month and thought it was doing something else.

The source of the 503 errors on startup is contile itself, [the handler returns them](https://github.com/mozilla-services/contile/blob/main/src/web/handlers.rs#L103) when the cache contains a `TilesState::Populating` entry. Requests set that state when they have a request to adM inflight, preventing subsequent requests from needlessly making the same redundant adM request until the original finishes (#248).

With that change we've ended only getting a pretty number of `AdmServerError`s (pretty much all are timeouts) -- they're all in sentry.

tools/test/integration_test.py
```diff
@@ -192,3 +200,12 @@ def test_bad_click_host(self, settings):
...
         assert list(names) == ["acme", "dunder mifflin"]
+
+    def test_aatimeout(self, settings):
+        # restart the test server with a timeout response
+        global SERVER
+        kill_process(SERVER)
+        setup_module(test_mode="TestTimeout")
+        url = "{root}/v1/tiles".format(root=settings.get("test_url"))
+        resp = requests.get(url, headers=default_headers(test=""))
+        assert resp.status_code == 204
```
pjenvey, line 211: With `AdmLoadError` going through the error handling path we might as well also `assert not resp.content`. I think `actix-web` doesn't render anything for 204s despite our error handler setting a JSON response body but let's test it.

tools/test/integration_test.py
```diff
@@ -192,3 +200,12 @@ def test_bad_click_host(self, settings):
         assert len(tiles) == 2
         names = map(lambda tile: tile.get("name").lower(), tiles)
         assert list(names) == ["acme", "dunder mifflin"]
+
+    def test_aatimeout(self, settings):
```
pjenvey, line 204: It seems like we don't actually run this test suite in CI?

src/adm/tiles.rs
```diff
@@ -172,39 +178,64 @@ pub async fn get_tiles(
...
+                    // We still want to track this as a server error later.
+                    //
+                    // TODO: Remove this after the shared cache is implemented.
+                    let mut err: HandlerError = if e.is_timeout()
+                        && Instant::now()
+                            .checked_duration_since(state.start_up)
+                            .unwrap_or_else(|| Duration::from_secs(0))
+                            <= Duration::from_secs(state.settings.adm_timeout)
+                    {
+                        HandlerErrorKind::AdmLoadError().into()
```
pjenvey, line 220: I'm not totally convinced we need all this because we don't seem to generate too many `AdmServerError`s AFAICT but I'm ok keeping it and seeing its actual usage. 

Though I'm thinking `AdmLoadError` should emit a metric instead of going to sentry (just needs a line added to `HandlerErrorKind::metric_label`) since we know exactly what causes it. I'd say the the same for `AdmServerError` (at least when `e.is_timeout()` which is the majority of its cases) but let's worry about that later.

pjenvey (changes requested):

tools/test/integration_test.py
jrconlin: no, we don't. There's a comment above noting that I don't know how to best test this using the current integration tests, but created this local test to verify that it's working.

src/adm/tiles.rs
jrconlin: The reason I think `AdmLoadError` should be a sentry event is because it's both significant and actionable, it's just that during the initial startup, there can be a lot of false events because we're swamping the sources. If we make it a metric, we're less inclined to take action. I am not going to make this a strong stand, though, so if we see different behaviors in the future, I'm find converting this to a metric.

pjenvey (approved):

PR #346, opened 2022-01-19
feat: Add support for multiple ADM endpoints

jrconlin:
ADM requires different endpoint data for mobile vs non-mobile tile
requests.

This patch also changes `AdmSettings` to `AdmFilterSettings` since
that's more explicit about what they are.

Issue: CONSVC-1562


src/adm/settings.rs
```diff
@@ -176,12 +177,56 @@ where
...
+    pub endpoint: String,
+}
+
+/// ADM Partner/Sub1/Endpoint.
+/// These change depending on the type of device requesting the tile.
+impl AdmPse {
+    /// Return the information for a mobile connection
+    pub fn as_mobile(settings: &Settings) -> Self {
+        let default = Self::as_default(settings);
+        AdmPse {
```
pjenvey, line 192: Instead of cloning the bits to create new `AdmPse`s every time we could construct them once at startup and store them on `ServerState`. I don't care much about these extra clones though so this as is is ok.

pjenvey (changes requested):

src/adm/settings.rs
```diff
@@ -176,12 +177,56 @@ where
...
+            sub1: settings.adm_mobile_sub1.clone().unwrap_or(default.sub1),
+            endpoint: settings
+                .adm_mobile_endpoint_url
+                .clone()
+                .unwrap_or(default.endpoint),
+        }
+    }
+
+    /// Return the information for a non-mobile connection
+    pub fn as_default(settings: &Settings) -> Self {
```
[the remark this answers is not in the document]
jrconlin: Thanks! I was debating the best way of dealing with this.
I also didn't really want to keep pumping stuff into ServerState if we didn't have to. This is probably a bit of my "if 2 than _n_" thinking, but i figure setting this pattern makes adding future ones a bit easier to deal with. Likewise, the clone I'm doing is just for the specific values, not the entire settings, so that should be reasonably lighter weight.

pjenvey (dismissed):

src/web/user_agent.rs
```diff
@@ -58,6 +58,12 @@ impl DeviceInfo {
         self.os_family == OsFamily::IOs && self.ff_version < 36
             || self.os_family != OsFamily::IOs && self.ff_version < 91
     }
+
+    /// Determine if the device is a mobile phone based on either the form factor or OS.
+    pub fn is_mobile(&self) -> bool {
+        self.form_factor == FormFactor::Phone
```
ncloudioj, line 64: @jrconlin Just confirmed with the mobile team, we do have mobile traffic on "tablets", and they wanted Contile to respond to tile requests from there. So let's include `FormFactor::Tablet` here as well.

pjenvey: Should we run this by adM as well?

ncloudioj: Yeah, we will confirm this with them in our next meeting.

src/web/user_agent.rs
```diff
@@ -58,12 +58,28 @@ impl DeviceInfo {
...
-    let wresult = Parser::new().parse(ua).unwrap_or_default();
-
+    let mut wresult = Parser::new().parse(ua).unwrap_or_default();
+
+    // NOTE: Firefox on iPads report back the "desktop" UA
+    // (e.g. `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4) AppleWebKit/605.1.15
+    //        (KHTML, like Gecko) Version/13.1 Safari/605.1.15)`
+    // therefore we have to accept that one. This does mean that we may presume
+    // that a mac safari UA is an iPad.
+    if wresult.name.to_lowercase().as_str() == "safari" && !ua.contains("firefox/") {
```
pjenvey, line 78:
```suggestion
    if wresult.name.to_lowercase() == "safari" && !ua.to_lowercase().contains("firefox/") {
```

pjenvey (changes requested):

jrconlin:
Now blocking on https://github.com/mozilla-services/contile-integration-tests/pull/147

hackebrot (changes requested):
Hi team! 👋🏻 

This is my understanding of the feature request:
- Send requests for tiles for mobile clients to a different partner API endpoint
- Make that URL configurable via `CONTILE_ADM_MOBILE_ENDPOINT_URL`

Unrelated to the partner mobile API, we’re also removing the deprecated settings `sub1` and `partner_id`. We're adding `adm_mobile_partner_id` and `adm_mobile_sub1` for the mobile API.

Does that sound correct to you?

src/adm/settings.rs
```diff
@@ -275,35 +324,22 @@ impl AdmSettings {
...
-                eprintln!(
-                    "{:?} is obsolete and will be removed. Please use {:?}",
-                    "partner_id", "adm_partner_id"
-                );
-            } else {
-                return Err(ConfigError::Message(format!("Missing argument {}", "$new")));
-            }
+            return Err(ConfigError::Message(format!(
+                "Missing argument {}",
+                "partner_id"
```
hackebrot, line 341: Should this be `adm_partner_id`?

jrconlin: yes, good catch.

jrconlin:
This is correct. 
This patch contains a number of fixes and changes which are products of the recent mobile endpoint changes. These normally get rolled up into the final commit message, but I will list them here for clarity.

* Add handling to send "mobile" devices to the new ADM mobile endpoint, with appropriate partner_id, and sub1 
* Add settings arguments for the new mobile specific arguments.
* Remove the obsolete `sub1` and  `partner_id` values (since we can no longer warn on their use.)
* "Correctly" identify firefox on iPad UserAgent string (which uses the default macos desktop UA per guidance from the mobile team and Apple)
* Remove the "position" field from outbound data since it is not used by User Agents and is complicating deployment. (Issue #347)
* Adjust integration testing to account for above.

hackebrot:
I'm currently working on a pull request for https://github.com/mozilla-services/contile-integration-tests/issues/148. 👨🏻‍💻 

I built the Contile Docker image based on this branch. I'm seeing an error with the existing Contile configuration here https://github.com/mozilla-services/contile-integration-tests/blob/main/docker-compose.yml.

```text
contile-integration-tests-contile-1  | Error: Missing argument adm_sub1
```

The Contile container exists with an error because the environment doesn't have `CONTILE_ADM_SUB1` set. I see you've added this here. What's the rationale for making this a required configuration? Do we set this for stage and prod?

integration-tests/volumes/client/scenarios_204.yml
```diff
@@ -1,6 +1,6 @@
 scenarios:
-  - name: success_204_No_Content_exluded_region
-    description: Test that Contile returns a 204 No Content for excluded regions
+  - name: success_200_No_Content_exluded_region
```
hackebrot, line 2: This change seems unrelated to the changes to support Mobile.

Did we change anything about the `CONTILE_EXCLUDED_COUNTRIES_200` configuration option?

integration-tests/volumes/partner/responses.yml
```diff
@@ -84,4 +84,17 @@ tablet:
     ios:
       status_code: 200
       headers: []
-      content: "hello world"
```
hackebrot, line 87: This response content is used for verifying that Contile correctly handles invalid tiles responses from the partner API in `error_tablet_ios_reqwest_error`. It is not supposed to match the schema of an actual response to a mobile client.

src/web/user_agent.rs
```diff
@@ -58,12 +58,28 @@ impl DeviceInfo {
...
 pub fn get_device_info(ua: &str) -> HandlerResult<DeviceInfo> {
-    let wresult = Parser::new().parse(ua).unwrap_or_default();
-
+    let mut wresult = Parser::new().parse(ua).unwrap_or_default();
+
+    // NOTE: Firefox on iPads report back the "desktop" UA
+    // (e.g. `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4) AppleWebKit/605.1.15
+    //        (KHTML, like Gecko) Version/13.1 Safari/605.1.15)`
+    // therefore we have to accept that one. This does mean that we may presume
+    // that a mac safari UA is an iPad.
```
hackebrot, line 77: What does that mean for Desktop clients using Safari on macOS?

Previously Contile would return a `403 Forbidden`. See the `error_invalid_user_agent` integration test.

hackebrot (dismissed):
Hi @jrconlin! 👋🏻 

I have added a few inline comments with questions about these changes.

Tomorrow I'll add support for a mobile endpoint to the test framework and help with updating the integration tests scenarios.

* https://github.com/mozilla-services/contile-integration-tests/issues/149

src/web/user_agent.rs
jrconlin: It means that if you have a UA string that matches the default Mac Desktop User Agent, we presume that you're on an iPad and provide tiles accordingly. I've adjusted the integration tests to account for that.

integration-tests/volumes/client/scenarios_204.yml
jrconlin: Yes. I believe that this was covered by https://mozilla-hub.atlassian.net/browse/CONSVC-1274

hackebrot: The `integration-tests/volumes/client/scenarios_204.yml` is used by `docker-compose.204.yml` which sets:

```yml
services:
  contile:
    environment:
      CONTILE_EXCLUDED_COUNTRIES_200: 0
```

Unless this has changed, we'd still expect a `204 No Content` with the configuration setting set to `0`.

integration-tests/volumes/partner/responses.yml
hackebrot: We needed a reliable way to return invalid tiles responses from the partner API in integration tests. While in prod iPads are expected to receive valid tiles data, we programmed the partner service in the integration tests to send an invalid response content for requests for form-factor `tablet` and os-family `ios`.

We control the User-Agent header in the `integration-tests/volumes/client/scenarios.yml` and know that Contile will receive an invalid response content from the partner and can verify that Contile handles that as expected.

pjenvey (approved):

pjenvey (approved):

hackebrot:
I merged `main` and resolved the merge conflicts.

pjenvey (dismissed):
LGTM just needs signoff from @hackebrot

jrconlin:
I'd also appreciate a drive by from @ncloudioj to make sure that the ADM calls are correct

src/web/user_agent.rs
```diff
@@ -58,12 +58,28 @@ impl DeviceInfo {
...
+    }
 }
 
 /// Parse a User-Agent header into a simplified `DeviceInfo`
 pub fn get_device_info(ua: &str) -> HandlerResult<DeviceInfo> {
-    let wresult = Parser::new().parse(ua).unwrap_or_default();
-
+    let mut wresult = Parser::new().parse(ua).unwrap_or_default();
+
+    // NOTE: Firefox on iPads report back the "desktop" UA
```
hackebrot, line 73: It's Safari on Desktop isn't it?

```suggestion
    // NOTE: Firefox on iPads report back the Safari "desktop" UA
```

src/web/test.rs
```diff
@@ -385,10 +384,8 @@ async fn basic_filtered() {
     // Ensure the tile order from adM is preserved
     let tile1 = &tiles[0];
     assert_eq!(tile1["name"], "Acme");
-    assert_eq!(tile1["position"], 0);
     let tile2 = &tiles[1];
     assert_eq!(tile2["name"], "Los Pollos Hermanos");
-    assert_eq!(tile2["position"], 2);
```
hackebrot, line 391: Do we have any unit test coverage for `position` not being included in the response content? We could invert this and the above change to check for that.

hackebrot (changes requested):
Hi @jrconlin @pjenvey! 👋🏻 

I've added my comments inline. I recommend using distinct values for partner parameters to make it easier to debug issues. Otherwise the changes look reasonable to me.

I appreciate you adding the comments to the integration test files. That's helpful! 👍🏻

src/web/test.rs
jrconlin: Don't the strict integration tests handle this? Adding a unit test for this feels like trying to prove a negative.

integration-tests/docker-compose.yml
```diff
@@ -26,8 +26,11 @@ services:
     environment:
       CONTILE_MAXMINDDB_LOC: /tmp/mmdb/GeoLite2-City-Test.mmdb
       CONTILE_ADM_ENDPOINT_URL: http://partner:5000/tilesp/desktop
+      CONTILE_ADM_MOBILE_ENDPOINT_URL: http://partner:5000/tilesp/mobile
       CONTILE_ADM_QUERY_TILE_COUNT: 5
       CONTILE_ADM_SETTINGS: /tmp/contile/adm_settings.json
+      CONTILE_ADM_SUB1: test
```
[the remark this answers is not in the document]
jrconlin: I suppose. Although, we aren't really checking the parameters in the mock parent script.

pjenvey (approved):

hackebrot (approved):
LGTM! 👍🏻 

Granted that @ncloudioj approves of these partner related changes, I'm happy for you to merge this pull request. 📱

Issue #348, opened 2022-01-25
Report error from image reader


jrconlin:
Currently not reporting the image library error.

Can be a Tag value to avoid cardinality.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1588)

PR #349, opened 2022-01-25
bug: report error from image library

jrconlin:
Closes: #348

### DIFF
--- a/src/adm/tiles.rs
+++ b/src/adm/tiles.rs
@@ -212,6 +212,7 @@ pub async fn get_tiles(
                     // We still want to track this as a server error later.
                     //
                     // TODO: Remove this after the shared cache is implemented.
+                    dbg!(&e);
                     let mut err: HandlerError = if e.is_timeout()
                         && Instant::now()
                             .checked_duration_since(state.start_up)
--- a/src/adm/tiles.rs
+++ b/src/adm/tiles.rs
@@ -136,7 +136,7 @@ pub async fn get_tiles(
     tags: &mut Tags,
     metrics: &Metrics,
     headers: Option<&HeaderMap>,
-) -> Result<TileResponse, HandlerError> {
+) -> HandlerResult<TileResponse> {
     let settings = &state.settings;
     let image_store = &state.img_store;
     let adm_url = Url::parse_with_params(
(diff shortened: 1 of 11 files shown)


ncloudioj:
@jrconlin When encountering this kind of error, is it possible for Contile to bail out rather than return an error code in the response? Doing so at least allows us to return some good tiles instead of nothing?

jrconlin:
This is currently failing tests because of the test suite change for mobile endpoints. #346 
Rather than spend time correcting for this change, I'm going to wait until after the multi-endpoint change lands, then fix things here.

src/server/img_storage.rs
```diff
@@ -190,9 +190,13 @@ impl StoreImage {
     pub fn meta(&self, image: &Bytes, fmt: ImageFormat) -> HandlerResult<ImageMetrics> {
         let mut reader = ImageReader::new(Cursor::new(image));
         reader.set_format(fmt);
-        let img = reader
-            .decode()
-            .map_err(|_| HandlerErrorKind::BadImage("Image unreadable"))?;
+        let img = reader.decode().map_err(|e| {
+            let mut tags = Tags::default();
+            tags.add_extra("error", &e.to_string());
```
pjenvey, line 195: I think both `url` and `format` are really worth adding as tags here too -- the easiest way to do that is to probably shuffle this method around a bit to have the same signature as `validate`.

pjenvey (changes requested):

src/server/img_storage.rs
jrconlin: Sigh. 
the change was included in the merge https://github.com/mozilla-services/contile/pull/349/commits/e767a61d02cb42123647db4604497a8e68c68ca3

src/adm/tiles.rs
```diff
@@ -212,6 +212,7 @@ pub async fn get_tiles(
                     // We still want to track this as a server error later.
                     //
                     // TODO: Remove this after the shared cache is implemented.
+                    dbg!(&e);
```
pjenvey, line 215: forgot to kill this

pjenvey (changes requested):

src/adm/tiles.rs
jrconlin: argh... hang on.

pjenvey (approved):

PR #351, opened 2022-02-02
feat: move cleanup of temp cache states into a ScopeGuard

### LINKED ISSUE #342: Improve tile handler's cache cleanup
In [PR #250](https://github.com/mozilla-services/contile/pull/250#discussion_r685472791) we complicated the tiles handler by adding a cleanup section to reset the tile cache state if any errors occurred. We ended up wrapping the work of the function where potential errors might occur in a closure, basically emulating a "try finally" block.

This was a quick fix with the intention that we improve it later. There's also a TODO (comment) that this solution doesn't handle the cache cleanup on potential panics.

The [scopeguard](https://github.com/bluss/scopeguard) crate offers what looks like an easy way to solve this instead, adding a Go-like defer macro that additionally works for panics.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1508)
┆epic: Contile production v1

pjenvey:
and kill TilesCache's Deref to dashmap, since we now only expose get and insert
outside of the cache module

Closes #342

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -670,6 +670,7 @@ dependencies = [
  "rand 0.8.4",
  "regex",
  "reqwest",
+ "scopeguard",
  "sentry",
  "sentry-backtrace",
  "serde 1.0.133",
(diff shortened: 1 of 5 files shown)


jrconlin (approved):
Looks like a cargo update should fix the audit err.

PR #352, opened 2022-02-02
chore: cargo update to resolve RUSTSEC-2022-0006

jrconlin:


pjenvey (approved):

hackebrot (approved):
I'm not familiar with each of the upgraded crates, but running `cargo update` sounds reasonable to me. Assuming it's common for Rust projects to follow semver, I see a few major version upgrades. As long we're fine with that, no concerns from me. 📦

jrconlin:
Rust library versioning is... odd. It generally follows SemVer (with the usual carve out for "0.*") How well that's followed is an exercise for the library author, though. We generally care most about the libraries we have direct reliance on, and those are specified in the `Cargo.toml` file. We tend to bind to the major.minor only, mostly because binding to the .release was felt as more problematic than helpful. There are a few libraries where we do this, to prevent accidental updates.

hackebrot:
Ok, sounds good. Thank you @jrconlin! 👍🏻

hackebrot:
You're right @mythmon. My mistake.

The way the Git diff was presented in the web UI, I assumed that `brotli-sys` was upgraded from `0.3.2` to `3.3.3`, but that's actually `brotli`. And `brotli2` from `0.3.2` to `2.3.2`, but that's actually `brotli-decompressor`.

Good to know that `cargo update` doesn't bump major versions.

PR #353, opened 2022-02-02
feat: Return JSON error messages

### LINKED ISSUE #177: Render actual JSON error responses
To aid in debugging we should return actual JSON error responses -- we currently return a payload of the error's status code, inherited from syncstorage-rs.

Technically this is json as we're sending `Content-Type: application/json` and a number, but this should be more along the lines of our other services's, e.g. megaphone's:

```
$ curl https://dev.megaphone.nonprod.cloudops.mozgcp.net/nonexistant
{"code":404,"errno":123,"error":"Not Found"}
```

Let's double check this won't affect the client but I believe it currently only cares about the http status code.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1089)
┆epic: Contile maintainablity enhancements

jrconlin:
Return more useful JSON error messages.
NOTE: We may not want to use the string versions of all errors we're
sending to sentry, so a "simplified" error message is generated.

Closes #177

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -149,7 +149,7 @@ impl HandlerErrorKind {
         match self {
             HandlerErrorKind::General(_) | HandlerErrorKind::Internal(_) => self.to_string(),
             HandlerErrorKind::Reqwest(_) => {
-                format!("An error occurred while trying to request data")
+                "An error occurred while trying to request data".to_string()
             }
             HandlerErrorKind::BadAdmResponse(_)
             | HandlerErrorKind::AdmServerError()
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -143,6 +144,29 @@ impl HandlerErrorKind {
     pub fn is_sentry_event(&self) -> bool {
         !matches!(self, HandlerErrorKind::InvalidUA)
     }
+
+    pub fn as_response_string(&self) -> String {
+        match self {
+            HandlerErrorKind::General(_) | HandlerErrorKind::Internal(_) => self.to_string(),
```
pjenvey, line 150: There were a couple places I thought might be using `Internal` while potentially exposing strings we don't want exposed, but it looks like they're all either called at startup (e.g. StoreImage::check_bucket`) or also the `AdmFilter` periodic updater that runs on its own thread.

Makes me wonder if we should move to using something else other than `Internal` for startupy errors in the future. Maybe `check_bucket` failing is really more of a `config::ConfigError` like we've started using elsewhere?

pjenvey (approved):

src/error.rs
jrconlin: `CloudStorage` could happen later, if the storage ever breaks for some random reason, but yeah, most of the others are either a `ConfigError` or something start-up related at least. Those probably don't need dedicated messages. 

We can add a ticket to review the errors we're recording at some later date.

Issue #354, opened 2022-02-09
Modify cloudstorage-rs to use a timeout 


data-sync-user:
Currently, {{cloud-storage}} does not specify a timeout value when making {{reqwest}} calls. This leads to high CPU if the connections fail for whatever reason which can lead to catastrophic outages.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1616)

data-sync-user:
➤ JR Conlin commented:

We currently include the cloud-storage-rs as an included github repo. We can modify that library to take an optional request::Client object. We should also file a ticket with the original library author so that other projects are not impacted by this bug.

data-sync-user:
➤ JR Conlin commented:

created https://github.com/ThouCheese/cloud-storage-rs/issues/99 ( https://github.com/ThouCheese/cloud-storage-rs/issues/99|smart-link ) to request access to the reqwest client. Looking at modifying our locally used fork of the code, or possibly using an alternate library implementation.

PR #355, opened 2022-02-10
bug: add timeouts for reqwest calls

jrconlin:
This uses new `mozilla-services/cloud-storage-rs *::*_with()` calls to
pass a `reqwest::Client` object with timeouts.

Closes: #354, CONSVC-1616

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -159,12 +159,9 @@ impl AdmFilter {
                     HandlerError::internal(&format!("Missing bucket Host {:?}", self.source))
                 })?
                 .to_string();
-            let obj = cloud_storage::Object::read_with(
-                &host,
-                bucket.path().trim_start_matches('/'),
-                req,
-            )
-            .await?;
+            let obj =
+                cloud_storage::Object::read_with(&host, bucket.path().trim_start_matches('/'), req)
+                    .await?;
             if let Some(updated) = self.last_updated {
                 // if the bucket is older than when we last checked, do nothing.
                 return Ok(updated <= obj.updated);
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -4,6 +4,7 @@ use std::{
     fmt::Debug,
     iter::FromIterator,
     sync::{Arc, RwLock},
+    time::Duration,
 };
 
 use actix_http::http::Uri;
@@ -58,7 +59,8 @@ pub struct AdmFilter {
     pub source: String,
     pub source_url: Option<url::Url>,
     pub last_updated: Option<chrono::DateTime<chrono::Utc>>,
-    pub refresh_rate: std::time::Duration,
+    pub refresh_rate: Duration,
+    pub timeout: Duration,
 }
 
 /// Parse &str into a `Url`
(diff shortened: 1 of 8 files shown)


ddurst:
Fixes CONSVC-1616

pjenvey (changes requested):

src/server/img_storage.rs
```diff
@@ -174,7 +174,11 @@ impl StoreImage {
         //
 
         // verify that the bucket can be read
-        let _content = Bucket::read(&settings.bucket_name)
+        let req = reqwest::Client::builder()
```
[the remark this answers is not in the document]
jrconlin: Thanks! missed that one was already there.

src/adm/filter.rs
```diff
@@ -150,14 +152,19 @@ impl AdmFilter {
             return Ok(false);
         }
         if let Some(bucket) = &self.source_url {
+            let req = reqwest::Client::builder().timeout(self.timeout).build()?;
```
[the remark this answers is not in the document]
jrconlin: I found a way to pass at least a clone instance down. That should ensure the timeout is handled.

Also replaced all the current `Client::new()` with `Client::builder()...` instances to make sure a timeout is present, even in test cases.

src/adm/filter.rs
```diff
@@ -144,7 +146,7 @@ impl AdmFilter {
     }
 
     /// check to see if the bucket has been modified since the last time we updated.
-    pub async fn requires_update(&self) -> HandlerResult<bool> {
+    pub async fn requires_update(&self, req: reqwest::Client) -> HandlerResult<bool> {
```
pjenvey, line 149: nit: Could avoid the clone in the updated closure
```suggestion
    pub async fn requires_update(&self, req: &reqwest::Client) -> HandlerResult<bool> {
```

src/adm/settings.rs
```diff
@@ -307,7 +308,11 @@ impl AdmFilterSettings {
             })?
             .to_string();
         let path = settings_bucket.path().trim_start_matches('/');
-        let contents = cloud_storage::Object::download(&bucket_name, path)
+        let req = reqwest::Client::builder()
```
pjenvey, line 311: forgot connect_timeout here as well, looks like you added it to all the tests where it doesn't matter much.

src/server/mod.rs
```diff
@@ -109,12 +109,12 @@ impl Server {
             raw_filter.update().await?
         }
         let filter = Arc::new(RwLock::new(raw_filter));
-        spawn_updater(&filter);
-        let tiles_cache = cache::TilesCache::new(TILES_CACHE_INITIAL_CAPACITY);
         let req = reqwest::Client::builder()
             .connect_timeout(Duration::from_secs(settings.connect_timeout))
```
pjenvey, line 113: let's add a `.timeout` call here, ensuring it's already set as a default everywhere we reuse this client. I think we'd 'overwrite' this timeout value in a couple places (like in `get_tiles`) but it doesn't really matter.

pjenvey (changes requested):

src/server/mod.rs
jrconlin: Ah, well, if we do that, we should specify both `connection_timeout` and `timeout` since they are used independently. I'll add the code to do that.

src/adm/settings.rs
jrconlin: Yeah, just wanted to be consistent.

src/server/mod.rs
ncloudioj: Agreed, looks like both `timeout` and `connection_timeout` are needed for all the Reqwest uses in Contile. The same applies to Merino once it needs to talk to external services at scale.

pjenvey (approved):
Looks good, just wanted to note (probably mostly for myself) that we also have an `adm_timeout` -- so the new `request_timeout` acts more as a global default and `adm_timeout` overwrites it for adm calls. Works for me

jrconlin:
Yeah, I know it feels like we're brimming with timeouts at this point, but I figure that partner connections will have their own criteria.

PR #359, opened 2022-02-22
feat: add a simple image cache to reduce image fetches

### LINKED ISSUE #357: Reduce Cloud Storage reads
Contile makes an excessive number of reads to its cloud storage bucket, particularly during the initial populating of its cache: for each non-empty cache entry being filled each Contile pod independently issues 1-2 reads (_adm_max_tiles_ = 2) to cloud storage, checking for the existence of the tile’s image icon.

As evidenced by the recent high number of DNS lookup failures of cloud storage’s hostname in logs, we’re making many redundant reads for a small number of unique icons.

A pretty simple way to reduce this traffic is for Contile to keep track of each icon uri it’s successfully read over the life of the program, skipping the uri’s image check entirely after 1 successful read.

One potential caveat here: does Contile need to take into account the lifetime/expiration of the image?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1629)

pjenvey:
we now refetch them only every 15 minutes (tiles_ttl) or sooner if we know of a
bucket level ttl

Closes #357

### DIFF
--- a/src/server/img_storage.rs
+++ b/src/server/img_storage.rs
@@ -106,8 +106,8 @@ impl Default for StorageSettings {
 
 /// Image storage container
 #[derive(Clone)]
-pub struct StoreImage {
-    // No `Default` stated for `StoreImage` because we *ALWAYS* want a timeout
+pub struct ImageStore {
+    // No `Default` stated for `ImageStore` because we *ALWAYS* want a timeout
     // for the `reqwest::Client`
     //
     // bucket isn't really needed here, since `Object` stores and manages itself,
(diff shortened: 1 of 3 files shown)


src/server/img_storage.rs
```diff
@@ -214,11 +237,17 @@ impl StoreImage {
...
     pub async fn store(&self, uri: &uri::Uri) -> HandlerResult<StoredImage> {
+        if let Some(image) = self.fetched_images.get(uri) {
+            if !image.expired() {
+                return Ok(image.clone());
+            }
+        }
         let (image, content_type) = self.fetch(uri).await?;
         let metrics = self.validate(uri, &image, &content_type).await?;
-        self.upload(image, &content_type, metrics).await
+        let image = self.upload(image, &content_type, metrics).await?;
```
jrconlin, line 248: nit: can we call this `stored_image` or something less confusing with the `image:Bytes` we had before?

jrconlin (approved):

jrconlin:
Also tempted to suggest we rename `StoreImage` to `ImageStorage` or something to avoid confusion with `StoredImage`

jrconlin (approved):

PR #364, opened 2022-02-25
Add Cloud Function based geo smoke-tests 🤖

hackebrot:
## Description

This pull request adds a new subdirectory with code for two Google Cloud Functions. We want to use these functions for a smoke-test suite to verify geolocation lookups in deployments.

The Cloud Functions are currently deployed manually in several different regions in the `contextual-services-test-eng` project.

## Issue(s)

CONSVC-1283


smoke-tests/tests/test_functions.py
```diff
@@ -0,0 +1,176 @@
...
+    """Start and terminate new subprocesses for the functions."""
+
+    p_clients = [
+        subprocess.Popen(
+            [
+                "functions-framework",
+                "--target",
+                client.target,
+                "--port",
+                str(client.port),
```
jrbenny35, line 85: Curious as to why you did it this was vs importing ```functions-framework```

smoke-tests/client/main.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+@dataclass
+class Error:
+    """Information about an error that occured."""
+
+    url: str
+    message: str
+    want: Any
+    got: Any
```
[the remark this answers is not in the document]
hackebrot: No, because this will also hold the response status code as an int.

We could change this to a `Union[Dict, int]`.

smoke-tests/tests/test_functions.py
hackebrot: It's the documented method of running integration tests. 📋 

See https://cloud.google.com/functions/docs/testing/test-http#integration_tests

smoke-tests/client/main.py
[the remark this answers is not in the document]
hackebrot: I pushed a commit to update the type annotations. 🤖

smoke-tests/tests/test_functions.py
hackebrot: I pushed a commit to document that in code.

jrconlin:
I'm not sure I understand the distinction between `client` and `runner`. Both look like they're doing the same thing. Is one just a dev artifact for now?

hackebrot:
> I'm not sure I understand the distinction between `client` and `runner`. Both look like they're doing the same thing. Is one just a dev artifact for now?

The `runner` triggers a set of `client` functions, which then run checks on the list of specified Contile environments. This way the Jenkins deployment pipeline only needs to send one HTTP request to trigger the smoke test suite and the `runner` takes care of aggregating the results.

There's a diagram in the **Software Architecture** section of the **Proposal: Automated smoke tests for Contile deployments** doc:

* `orchestrator` in the doc is `runner` here
* `app` in the doc is `client` here

Hope this helps!

jrconlin (approved):

pjenvey (approved):

hackebrot:
Thanks for the reviews, team! 🚀

Issue #377, opened 2022-03-24
Add filter for `image_url` elements


data-sync-user:
We currently do not impose a filter on the hosts we allow for `image_url`. This may require communicating with ADM to get the list of per-partner hosts. 

I don’t believe that we have any path requirements for these.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1716)
┆epic: Contile reliability enhancements

data-sync-user:
➤ JR Conlin commented:

PR https://github.com/mozilla-services/contile/pull/378 ( https://github.com/mozilla-services/contile/pull/378|smart-link )

PR #378, opened 2022-03-24
bug: add check for image_hosts

jrconlin:
Closes: #377

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -349,7 +349,7 @@ impl AdmFilter {
     /// Check the image URL to see if it's valid.
     ///
     /// This extends `filter_and_process`
-    fn check_img_hosts(
+    fn check_image_hosts(
         &self,
         filter: &AdmAdvertiserFilterSettings,
         tile: &mut AdmTile,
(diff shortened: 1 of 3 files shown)


src/adm/settings.rs
```diff
@@ -129,6 +129,12 @@ pub struct AdmAdvertiserFilterSettings {
         default
     )]
     pub(crate) click_hosts: Vec<Vec<String>>,
+    #[serde(
+        deserialize_with = "deserialize_hosts",
+        serialize_with = "serialize_hosts",
+        default
+    )]
+    pub(crate) img_hosts: Vec<Vec<String>>,
```
pjenvey, line 137: nit: not terribly important but might as well match the field name in the json's prefix
```suggestion
    pub(crate) image_hosts: Vec<Vec<String>>,
```

src/adm/filter.rs
```diff
@@ -430,7 +456,12 @@ impl AdmFilter {
                     self.report(&e, tags);
                     return None;
                 }
-
+                if let Err(e) = self.check_image_hosts(img_filter, &mut tile, tags) {
+                    trace!("Rejecting tile: bad img");
+                    metrics.incr_with_tags("filter.adm.err.invalid_img_host", Some(tags));
```
pjenvey, line 461: Sorry to be pedantic but one more here, especially because we already have a few metrics with the word "image" in it.
```suggestion
                    metrics.incr_with_tags("filter.adm.err.invalid_image_host", Some(tags));
```

pjenvey (changes requested):

src/adm/filter.rs
jrconlin: Nope, good catch! I had done variable refactors for the others and they missed this.

pjenvey (approved):

PR #382, opened 2022-05-04
fix: Add get_id_token function and Authorization header

dlactin:
Client cloud functions are configured to require a valid ID token
associated with a service account that has the invoke permission.
This change adds a method to request an ID token and adds it as
a header to the request invoking the client functions.

PR #383, opened 2022-05-06
Update geolocation smoke-test runner

hackebrot:
## Description

@dlactin figured out why the runner function fails to trigger the client functions.

This pull request from the main repo replaces #382 because we don't run CI for forks, which means the other pull request is blocked from merging.


pjenvey (approved):
Thanks, I often cheat and disable the checks temporarily for those PRs.

PR #387, opened 2022-05-11
feat: upgrade to actix-web 4

### LINKED ISSUE #386: Upgrade to actix-web 4
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1783)

pjenvey:
- move to the latest cloud-storage (w/ a new API)
- upgrade cadence, requiring wrapping of its Client in Arc
- fix the 404 response to use the standard json payload

Closes #386

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,44 +16,41 @@ debug = 1
 
 [dependencies]
 actix-cors = "0.6"
-actix-web = "4"
+actix-web = { version = "4", default_features = false, features = ["macros"] }
 actix-web-location = { version = "0.7", features = ["actix-web-v4", "maxmind", "cadence"] }
 async-trait = "0.1"
 backtrace = "0.3"
 base64 = "0.13"
-blake3 = "1.0"
+blake3 = "1"
 bytes = "1"
 cadence = "0.29"
 chrono = "0.4"
 docopt = "1.1"
 cloud-storage = { git = "https://github.com/mozilla-services/cloud-storage-rs", branch = "release/0.11.1-client-builder-and-params" }
-config = "0.11"
-dashmap = "4.0.2"
+config = "0.13"
+dashmap = "5.3"
 futures = "0.3"
 gethostname = "0.2.1"
 hex = "0.4"
-hostname = "0.3"
 image = "0.24"
 lazy_static = "1.4"
 log = { version = "0.4", features = ["max_level_trace", "release_max_level_info"] }
 rand ="0.8"
-regex = "1.4"
+regex = "1"
 reqwest = { version = "0.11", features = ["json"] }
-serde = "1.0"
-# pin to 0.19 (until our onpremise is upgraded):
-# https://github.com/getsentry/sentry-rust/issues/277
-sentry = "0.19"
-sentry-backtrace = "0.19"
-serde_json = "1.0"
-scopeguard = "1.1.0"
+serde = "1"
+sentry = "0.25"
+sentry-backtrace = "0.25"
+serde_json = "1"
+scopeguard = "1.1"
 slog = { version = "2.7", features = ["max_level_trace", "release_max_level_info", "dynamic-keys"] }
-slog-async = "2.6"
+slog-async = "2.7"
 slog-envlogger = "2.2.0"
 slog-mozlog-json = "0.1"
 slog-scope = "4.4"
 slog-stdlog = "4.1"
-slog-term = "2.7"
-thiserror = "1.0"
+slog-term = "2"
+thiserror = "1"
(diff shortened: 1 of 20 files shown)


src/web/handlers.rs
```diff
@@ -193,7 +193,7 @@ fn content_response(content: &cache::TilesContent) -> HttpResponse {
     match content {
         cache::TilesContent::Json(json) => HttpResponse::Ok()
             .content_type("application/json")
-            .body(json),
+            .body(json.to_owned()),
```
pjenvey, line 196: I'll note that actix-web has always internally made a copy of data like `&str` here for response bodies. actix-web 4's API just exposes that to the user, requiring them to make the copy. This makes sense with responses being used in asynchronously, a lifetime would complicate that.

However a `&'static str` can be easily used asynchronously so it supports that and we utilize it w/ `body(EMPTY_TILES.as_str())` (no copying) above.

Cargo.toml
```diff
@@ -15,20 +15,18 @@ edition = "2021"
...
 blake3 = "1.0"
-bytes = "1.0"
-cadence = "0.26"
+bytes = "1"
+cadence = "0.29"
 chrono = "0.4"
 docopt = "1.1"
-cloud-storage = { git = "https://github.com/mozilla-services/cloud-storage-rs", branch = "release/0.6.2-create_with_params" } # 0.7+ includes request 0.11, tokio 1.4
+cloud-storage = { git = "https://github.com/mozilla-services/cloud-storage-rs", branch = "release/0.11.1-client-builder-and-params" }
 config = "0.11"
```
jrconlin, line 30: Didn't want to bump this to 0.13?

Cargo.toml
```diff
@@ -15,20 +15,18 @@ edition = "2021"
...
 dashmap = "4.0.2"
```
jrconlin, line 31: Looks like bumping this to 5.3 has no code impact

jrconlin (commented):
Do we also want to bump sentry up to 0.25?

jrconlin:
Do we also want to bump sentry up to 0.25?

Cargo.toml
pjenvey: I was planning to upgrade more things in another PR (like sentry)

jrconlin (approved):

src/adm/settings.rs
```diff
@@ -320,7 +320,13 @@ impl AdmFilterSettings {
             .timeout(request_timeout)
             .build()
             .map_err(|e| ConfigError::Message(e.to_string()))?;
-        let contents = cloud_storage::Object::download_with(&bucket_name, path, &req)
+        let storage_client = cloud_storage::Client::builder()
```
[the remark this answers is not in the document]
pjenvey: While collapsing this I ended up with a small refactor which ends up sharing the cloud storage Client a bit more

src/server/img_storage.rs
```diff
@@ -495,14 +507,21 @@ mod tests {
     fn test_store() -> ImageStore {
         let settings = test_storage_settings();
         let timeout = std::time::Duration::from_secs(settings.request_timeout);
+        let req = reqwest::Client::builder()
```
jrconlin, line 510: nit: tempted to rename this `client` instead of `req`.

src/server/mod.rs
```diff
@@ -95,28 +99,31 @@ macro_rules! build_app {
...
     pub async fn with_settings(mut settings: Settings) -> Result<dev::Server, HandlerError> {
-        let metrics = metrics_from_opts(&settings)?;
-        let mut raw_filter = HandlerResult::<AdmFilter>::from(&mut settings)?;
-        // try to update from the bucket if possible.
-        if raw_filter.is_cloud() {
-            raw_filter.update().await?
-        }
-        let filter = Arc::new(RwLock::new(raw_filter));
+        let metrics = Arc::new(metrics_from_opts(&settings)?);
         let req = reqwest::Client::builder()
```
jrconlin, line 110: nit: tempted to rename this `client`.
We're building a `reqwest::Client`, that is passed to `.client()` and stored as `reqwest_client:`

jrconlin (approved):

src/server/mod.rs
pjenvey: Agreed we can clean up this naming, I'll try to tackle it on later PR. I probably wouldn't use `client` now that we also have a `storage_client: cloud_storage::Client`. Maybe `reqwest` and `cloud_storage`?

pjenvey:
I'll wait for #385 before merging this

jrconlin (approved):

Issue #284, opened 2021-09-22
Ensure non-live Countries do not see tiles


data-sync-user:
Follow up to the 9/16 incident, all countries temporarily received the fallback set of tiles.

After the incident was resolved these countries began receiving 204 empty responses. Desktop appears to [treat 204s as a noop, reverting back to its existing tile set](https://hg.mozilla.org/integration/autoland/file/7ab576fad46b5017a362797ae20ac8e4eb761622/browser/components/newtab/lib/TopSitesFeed.jsm#l165) which means these countries' clients are potentially still seeing tiles they shouldn't at least until they restart Firefox.

If that's the case we should be able to workaround this behavior by returning them empty tile sets w/ a 200 OK instead.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/CONSVC-1266)

PR #446, opened 2022-09-02
feat: integrate shepherd output for contile

tiftran:
main things that in the changes to accommodate the shepherd output:

click_hosts, impression_hosts have been removed from advertiser specific level
DEFAULTS have been moved to their own settings config env
at the advertiser specific level is where the country info lies now

Closes [CONSVC-1807](https://mozilla-hub.atlassian.net/jira/software/c/projects/CONSVC/issues/CONSVC-1807)


src/adm/settings.rs
```diff
@@ -408,45 +514,73 @@ impl From<&mut Settings> for HandlerResult<AdmFilter> {
...
+            if let Ok(mut f) = File::open(&settings.adm_settings) {
+                settings_str = f
+                    .read_to_string(&mut settings.adm_settings)
+                    .map_err(|e| {
+                        HandlerError::internal(&format!(
+                            "Could not read {}: {:?}",
+                            settings.adm_settings, e
+                        ))
+                    })?
+                    .to_string();
```
tiftran, line 554: this feels jank... im not sure how to not make it feel less jank

src/adm/settings.rs
```diff
@@ -112,41 +126,20 @@ pub struct PathFilter {
...
-    pub(crate) image_hosts: Vec<Vec<String>>,
-    /// valid position for the tile
-    pub(crate) position: Option<u8>,
-    /// Optional set of valid countries for the tile (e.g ["US", "GB"])
-    /// TODO: could support country + subdivision, e.g. "USOK"
-    #[serde(default)]
-    pub(crate) include_regions: Vec<String>,
-    pub(crate) ignore_advertisers: Option<Vec<String>>,
-    pub(crate) ignore_dmas: Option<Vec<u8>>,
+    // TODO: handle country and path parsing.
```
jrconlin, line 129: Can probably remove this TODO

src/adm/settings.rs
```diff
@@ -244,62 +265,171 @@ impl Serialize for AdmFilterSettings {
...
+        // first level `ADVERTISER: AdvertiserFilterSettings`
+        for (advertiser, value) in parsed {
+            // second level `COUNTRY:Vec<AdvertiserUrlFilter> | "delete": bool`
+            let mut countries: HashMap<String, Vec<AdvertiserUrlFilter>> = HashMap::new();
+            let mut delete = false;
+            for (key, value) in value.as_object().ok_or_else(|| {
+                ConfigError::Message(format!("Invalid advertiser info {}", &advertiser))
+            })? {
+                // Delete allows us to delete this advertiser. It may appear at the same
+                // level as the country listsings.
```
jrconlin, line 327: s/listsings/listings/

src/adm/settings.rs
```diff
@@ -244,62 +265,171 @@ impl Serialize for AdmFilterSettings {
...
+                if key.to_lowercase() == "delete" {
+                    delete = value.as_bool().unwrap_or_default();
+                    continue;
+                }
+                // if it's not a `delete`, than it's part of the
+                // list of per country advertiser filters.
+                //*
```
jrconlin, line 334: Let's remove the comment block bits here. We use this code.

src/adm/settings.rs
```diff
@@ -319,11 +449,10 @@ impl AdmFilterSettings {
             .download(&bucket_name, path)
             .await
             .map_err(|e| ConfigError::Message(format!("Could not download settings: {:?}", e)))?;
-        let mut reply =
-            AdmFilterSettings::try_from(String::from_utf8(contents).map_err(|e| {
+        let reply =
```
jrconlin, line 452: kinda wondering if we could collapse this into the `OK()`

e.g.
``` 
        Ok(AdmFilter::advertisers_from_string(&String::from_utf8(contents).map_err(|e| {
                ConfigError::Message(format!("Could not read ADM Settings: {:?}", e))
            })?)?)
```

src/adm/settings.rs
```diff
@@ -408,45 +521,68 @@ impl From<&mut Settings> for HandlerResult<AdmFilter> {
             .clone()
             .unwrap_or_else(|| "[]".to_owned())
             .to_lowercase();
-        let mut all_include_regions = HashSet::new();
+        // No longer needed since each advertiser has it's own set of countries.
+        // let mut all_include_regions = HashSet::new();
```
jrconlin, line 525: Since it's no longer needed, let's drop this.

src/adm/settings.rs
```diff
@@ -455,10 +591,10 @@ impl From<&mut Settings> for HandlerResult<AdmFilter> {
 mod tests {
     use std::env;
 
-    use serde_json::json;
+    //use serde_json::json;
 
     use super::*;
-    use crate::web::test::adm_settings;
+    //use crate::web::test::adm_settings;
```
jrconlin, line 597: And remove the commented out `// use` statements.

src/web/test.rs
```diff
@@ -108,11 +120,12 @@ impl MockAdm {
...
     async fn params(&mut self) -> HashMap<String, String> {
         let query_string = self.request_rx.next().await.expect("No request_rx result");
-        Url::parse(&format!("{}{}", self.endpoint_url, query_string))
+        let bits = Url::parse(&format!("{}?{}", self.endpoint_url, query_string))
             .expect("Couldn't parse request_rx result")
             .query_pairs()
             .into_owned()
-            .collect()
+            .collect();
+        bits
```
jrconlin, line 128: since we're no longer debugging this, let's just return 
```
Url::parse(&format!("{}?{}", self.endpoint_url, query_string))
            .expect("Couldn't parse request_rx result")
            .query_pairs()
            .into_owned()
            .collect()
```

test-engineering/contract-tests/volumes/client/scenarios.yml
```diff
@@ -231,297 +231,297 @@ scenarios:
...
-                      value: 'desktop'
-                    - name: os-family
-                      value: 'windows'
-                    - name: v
-                      value: '1.0'
-                    - name: out
-                      value: 'json'
-                    - name: results
-                      value: '5'
+#  - name: legacy_image_filter
```
jrconlin, line 234: We should drop this or at least add a comment block indicating that we no longer need to do image filter tests, since they're now part of the default configuration. 

Make sure to also flag context-services-test-eng as a reviewer since we're modifying the tests.

test-engineering/contract-tests/README.md
```diff
@@ -17,9 +17,7 @@ the individual components of the suite.
 To run the contract tests locally, execute the following from the repository root:
 
 ```text
-docker compose \ 
--f test-engineering\contract-tests\docker-compose.yml \ 
-up --abort-on-container-exit --build
+u
```
jrconlin, line 20: Guessing this is a typo? Probably restore what was there.

jrconlin (changes requested):
A few minor fix ups, but looks good. Thanks!

src/adm/filter.rs
```diff
@@ -45,22 +45,22 @@ lazy_static! {
 /// In addition there is a special `DEFAULT` value which is a filter
```
jrconlin, line 45: Let's delete references to the `DEFAULT` since it's no longer present.

jrconlin (changes requested):
Looks pretty good. A few minor fixups to remove some older comments and collapse some code.

test-engineering/contract-tests/README.md
tiftran: eeek oops

Trinaa (changes requested):
I believe that there are contract tests being excluded, for instance `success_tiles_cache_expired`, that are intended to test functionality unrelated to this change. Perhaps we can discuss and adjust the scenarios accordingly?

test-engineering/contract-tests/volumes/client/scenarios.yml
tiftran: eek, lemme go through those tests and uncomment accordinly first, I initially commented out to help me test locally.

jrconlin:
@Trinaa: We commented out those tests initially because we knew there would be some work getting them back online. We changed around a number of the configs to fit the way that Contile actually works, rather than the wild guesses we made initially. (e.g. the Click, Image, and Impression URLs don't really change, since we own those, so they were moved out of ADM_Settings and into ADM_Defaults. We just forgot to turn those tests back on. (oops.)

I've got most of the tests working now, but there are a couple that i'm hitting that I'm not sure how to fix. One of them is the one that's currently failing. I think it's mostly a matter of trying to mesh up the inputs and responses. If you have any suggestions, I'm all ears.

test-engineering/contract-tests/docker-compose.yml
```diff
@@ -30,9 +30,23 @@ services:
...
+        "impression_hosts": [
+            "example.com",
+            "example.org",
+            "dunbroch.co.uk"
+        ]
+      }'
       CONTILE_ADM_SUB1: sub1_test
       CONTILE_ADM_PARTNER_ID: partner_id_test
       CONTILE_ADM_HAS_LEGACY_IMAGE: '["Example ORG","Example COM"]'
+      CONTILE_ADM_DEFAULTS: '{"image_hosts":["example.com","example.org"],"impression_hosts":["example.com","example.org"],"click_hosts":["example.com","example.org"]}'
```
Trinaa, line 49: This is a duplicate environment variable.

test-engineering/contract-tests/docker-compose.yml
```diff
@@ -30,6 +30,24 @@ services:
...
+        "click_hosts": [
+            "example.com",
+            "example.org",
+            "dunbroch.co.uk"
+        ],
+        "impression_hosts": [
+            "example.com",
+            "example.org",
+            "dunbroch.co.uk"
+        ],
```
Trinaa, line 44: **nit:** indentation

```suggestion
        "click_hosts": [
          "example.com",
          "example.org",
          "dunbroch.co.uk"
        ],
        "impression_hosts": [
          "example.com",
          "example.org",
          "dunbroch.co.uk"
        ],
```

test-engineering/contract-tests/volumes/contile/adm_settings.json
```diff
@@ -1,104 +1,186 @@
...
-            "DE",
-            "FR",
-            "ES",
-            "IN",
-            "MX",
-            "CN"
-        ]
-    }
+  "Example COM": {
+    "US": [
```
Trinaa, line 3: **suggestion:** Would it make sense to order regions alphabetically?

Trinaa (commented):
@tiftran @jrconlin 
What do you think about adding a contract test for the new `delete` advertiser key field (could be a subsequent task depending on feasibility and effort)?

tiftran:
@Trinaa @jrconlin I think I want to opt to add the contract test in another PR considering this one is so big already

Trinaa (approved):
@tiftran

Contract test changes LGTM 🚀 . Let me know if/how I can support the contract test addition in subsequent PR.

jrconlin (approved):

jrconlin (approved):

src/adm/tiles.rs
```diff
@@ -248,17 +249,22 @@ pub async fn get_tiles(
...
-        response
-            .tiles
-            .into_iter()
-            .filter_map(|tile| {
-                filter.filter_and_process(tile, location, &device_info, tags, metrics)
-            })
-            .take(settings.adm_max_tiles as usize)
-            .collect()
-    };
+    let mut filtered: Vec<Tile> = Vec::new();
```
pjenvey, line 252: Is this change necessary? Not that it's terribly important with `adm_max_tiles` being small but previously it only acquired the read lock once.

test-engineering/contract-tests/volumes/contile/adm_settings.json
```diff
@@ -1,104 +1,186 @@
...
-            "CA",
-            "DE",
-            "FR",
-            "ES",
-            "IN",
-            "MX",
-            "CN"
-        ]
-    }
+  "Example COM": {
```
pjenvey, line 2: Sorry I should have suggested this much earlier, but how about adding another level of nesting to these json settings? It would make adding extra buckets of data later easier if needed.

```json
{
    "adm_advertisers": {
        "Example COM": {
             ...
        }
    },
    "last_modified": "2022-10-17 21:50:00 UTC"
}
```

E.g. a timestamp from shepherd (granted we could also use GCS's file timestamps for that)

src/adm/tiles.rs
jrconlin: IIRC, the problem is that `filter_and_process()` now returns a `Result<Option<...>>`. I can't just call `?` on it otherwise I'll lose the error. That said, I can break `filter` out of the loop fairly easily.

test-engineering/contract-tests/volumes/contile/adm_settings.json
jrconlin: Let's make that a future feature, since we'll have to modify Shepherd output as well.

jrconlin:
Having talked with @pjenvey I think there are a few changes we should definitely consider before this gets landed. Sadly, these changes will also require changes to the Shepherd UI.

1) we should move the advertisers into a sub block (see [this comment](https://github.com/mozilla-services/contile/pull/446#pullrequestreview-1144845467)] 
2) Let's drop the `delete` flag. If an advertiser is no longer to be recognized, it should just not be part of the data. This will also mean we can use proper `serde` to read data in, so we can drop a lot of [this code](https://github.com/mozilla-services/contile/pull/446/files#diff-34387244eabf61c959f3a7e535ca04a551c779b9ee645d2e93303c0020f5a881)

If you run out of rude things to say about me, feel free to ask. I'll provide an endless supply of material you can use to insult me with.

jrconlin (changes requested):
See [this comment](https://github.com/mozilla-services/contile/pull/446#issuecomment-1282845890)

src/adm/tiles.rs
pjenvey: My fault, I missed this in the diff. You could collect into `Vec<Result<..>>` but that doesn't work well with the `take`ing, so this change makes sense.

tiftran:
@jrconlin @pjenvey just wanted to get a check on removing delete and adding the sub level before moving onto adding back the include_regions bit?

jrconlin:
Yes, we want to remove the "delete" option (which should also GREATLY simplify things, since we should be able to remove [the custom parser code](https://github.com/mozilla-services/contile/pull/446/files#diff-34387244eabf61c959f3a7e535ca04a551c779b9ee645d2e93303c0020f5a881R291) and just use simple `serde` rules. We also want to put the `adm_filters` under their own section. I don't think we really care what they should be called, but basically the Company Name and filter info should not be at the top level.

tiftran:
@jrconlin @pjenvey so i _think_ i added the sub level, removed the manually parsing bits. Was hoping if you could look over my new commits, before i start adding back the include regions stuff? 🥺

src/adm/filter.rs
```diff
@@ -380,14 +380,10 @@ impl AdmFilter {
     ) -> HandlerResult<Option<Tile>> {
         // Use strict matching for now, eventually, we may want to use backwards expanding domain
         // searches, (.e.g "xyz.example.com" would match "example.com")
-        match self
-            .advertiser_filters
-            .adm_advertisers
-            .get(&tile.name.to_lowercase())
-        {
+        match self.advertiser_filters.adm_advertisers.get(&tile.name) {
```
jrconlin: This does mean that "Amazon" and "amazon" are different companies. Do we want that?

src/adm/settings.rs
```diff
@@ -230,76 +237,131 @@ impl AdmPse {
...
 }
 
 /// Create AdmSettings from a string serialized JSON format
-impl TryFrom<String> for AdmFilterSettings {
-    type Error = ConfigError;
+impl AdmFilter {
+    /// Parse a JSON string containing the ADM settings. These will be generated by shepherd and
+    /// would have a format similar to the following:
+    /// ```json
+    /// {
```
jrconlin, line 315: Let's update this to reflect the new format.

src/adm/settings.rs
```diff
@@ -333,70 +395,54 @@ impl AdmFilterSettings {
...
-        AdmFilterSettings::try_from(settings_str)
-    }
-}
 
 /// Construct the AdmFilter from the provided settings.
 ///
 /// This uses a JSON construct of settings, e.g.
 /// ```javascript
 /// /* for the Example Co advertiser... */
 /// {"Example": {
```
jrconlin, line 404: Again, let's make this comment match what we're doing, so add in the parent `adm_advertisers` key.

src/adm/settings.rs
```diff
@@ -477,46 +553,90 @@ mod tests {
...
-        let filter = HandlerResult::<AdmFilter>::from(&mut settings).unwrap();
-        assert!(
-            filter.all_include_regions
-                == vec!["US", "MX"]
-                    .into_iter()
-                    .map(ToOwned::to_owned)
-                    .collect()
-        );
+    pub fn test_valid_path_filters() {
+        let adm_settings = r#"{"adm_advertisers":{"test-adv": {
```
jrconlin, line 561: probably want to pretty this so it's more readable.

adm_settings_test.json
```diff
@@ -1,23 +1,5 @@
...
-        "FR",
-        "ES",
-        "IN",
-        "MX",
-        "CN"
-    ]
-},
-"Acme": {
-    "advertiser_urls": [
+{"Acme": {
```
jrconlin, line 1: Shouldn't this include the `adm_advertisers` parent?

src/adm/filter.rs
jrconlin: (I think we want to keep the `name.to_lowercase()` because we're normalizing the names when we add them to `adm_advertisers`.

tiftran: huh i thought i added that back... will re add,

jrconlin:
FWIW, I've had stuff like that get dropped because of a merge. I chalk it
up as "git being helpful".  Might be what happened here.

src/adm/filter.rs
```diff
@@ -181,28 +187,25 @@ impl AdmFilter {
...
+                        HandlerError::internal(&format!(
+                            "Invalid bucket data in {:?}: {:?}",
+                            self.source, e
+                        ))
+                    })?;
+            self.advertiser_filters.adm_advertisers.clear();
+            for (adv, setting) in advertiser_filters.adm_advertisers {
+                for country in setting.keys() {
+                    if !self.defaults.include_regions.contains(country) {
+                        self.defaults.include_regions.push(country.clone());
```
tiftran, line 203: wasn't sure if this was the way to go about adding include_regions

src/adm/filter.rs
jrconlin: Are we not indexing by lowercase? If so, we probably want to remove the lowercase we do in `Deserialize<'de> for AdmAdvertiserSettings`. 

Probably should add a unit test somewhere to ensure that we don't trip on this in the future.

jrconlin (changes requested):

src/adm/filter.rs
tiftran: Screen Shot 2022-10-26 at 11 49 04 AM
this is the current diff i see?

jrconlin (approved):

src/adm/filter.rs
pjenvey: I don't think `include_regions` belongs in `AdmDefaults`, it's a bit confusing there since it should only be generated by contile. Despite the previous `DEFAULT` key supporting an `include_regions` value (which was a bit different) we didn't really need it and especially don't with shepherd's UI and the new per region advertiser settings, so it was good that you originally got rid of it.

Let's keep it as `AdmFilter::all_include_regions` (`all_`'s helpful to distinguish it from the `include_regions` settings) and as a `HashSet`. I think you can pretty much revert to most of the previous code, and this `update()` could easily `.clear()` and rebuild it.

cc @jrconlin we can probably kill `NoTilesForCountry` as well?

src/adm/filter.rs
```diff
@@ -234,26 +236,23 @@ impl AdmFilter {
...
-                continue;
-            }
-
-            if let Some(ref paths) = filter.paths {
+        for filter in filters {
+            if host == filter.host {
+                let paths = &filter
+                    .paths
+                    .clone()
+                    .unwrap_or_else(|| [PathFilter::default()].to_vec());
```
pjenvey, line 244: Nit: you could avoid this clone here by keeping a static of the default:

```suggestion
                    .as_ref()
                    .unwrap_or_else(|| &DEFAULT_PATH_FILTER);
```

with this in the `lazy_static` block:
```rust
static ref DEFAULT_PATH_FILTER: Vec<PathFilter> = vec![PathFilter::default()];
```

pjenvey, line 244: Nit: you could avoid this clone here by keeping a static of the default:

```suggestion
                let paths = filter
                    .paths
                    .as_ref()
                    .unwrap_or_else(|| &DEFAULT_PATH_FILTER);
```

with this in the `lazy_static` block:

```rust
static ref DEFAULT_PATH_FILTER: Vec<PathFilter> = vec![PathFilter::default()];
```

src/adm/filter.rs
jrconlin: I think we need to keep `NoTilesForCountry`. We have logic still in `filter_and_process` to optionally return a 200 for tiles that are in excluded countries (since those default to return 204), and `NoTilesForCountry` allows us to return that in `error.rs`.

pjenvey: I see what you're getting at: The `settings.excluded_countries_200` flag works around [what's really a bug in the client](https://github.com/mozilla-services/contile/issues/284) where it treats our 204 EMPTY response as a "noop, reverting back to the existing tile set it has", so this setting makes us return a 200 OK instead, forcing the client to use an empty tile set instead.

You're right in that we've only applied this setting to that one code path and we're still rendering 204s in other cases, with `NoTilesForCountry` covering one of them.

However there is a bug with it, it basically "short circuits" the filtering process, I've made a test on another branch which it causes to fail: https://github.com/mozilla-services/contile/compare/shepherd-output...shepherd-output-filter-bug (we're returning empty tiles there instead of 1)

I think it makes sense to consider the other 204s we're still rendering as a bug (separate from this PR) but I think the right fix is passing this setting along to all the places rendering 204s where they can take it into account.

pjenvey (approved):
I'd prefer passing the `excluded_countries_200` setting to the render methods instead of storing this in the cache but no big deal, can always change it later.

Thanks @tiftran & @jrconlin!

src/adm/filter.rs
```diff
@@ -381,34 +380,24 @@ impl AdmFilter {
         device_info: &DeviceInfo,
         tags: &mut Tags,
         metrics: &Metrics,
-    ) -> Option<Tile> {
+    ) -> HandlerResult<Option<Tile>> {
```
pjenvey, line 383: This might not need to return a Result any longer (I'm guessing it was added for NoTilesForCountry)?

pjenvey (approved):

PR #454, opened 2022-09-23
chore: Quarterly dependency updates

jrconlin:


ethowitz:
Should the rust version in the circleci images be updated to 1.64.0 as well? It looks like they're pinned to 1.62.0 🤔

pjenvey:
> Should the rust version in the circleci images be updated to 1.64.0 as well? It looks like they're pinned to 1.62.0 🤔

Yup. Let's also improve the comment above the Dockerfile's version line to make it more clear that .circleci/config.yml should be bumped alongside it.

pjenvey:
FYI as to why there were clippy CI failures this morning despite ci's "pinning": due to a [bug where we unpinned](https://github.com/mozilla-services/contile/pull/455)

ncloudioj (approved):

PR #479, opened 2022-11-09
fix: 200 OK should always include empty tiles set

tiftran:


src/server/cache.rs
```diff
@@ -246,7 +247,11 @@ impl Tiles {
                 if cache_control_header {
                     builder.insert_header(self.cache_control_header());
                 }
-                builder.finish()
+                if self.always_ok {
+                    builder.body(EMPTY_TILES.as_str())
```
tiftran, line 251: 🙈  this is really jank, but i wasn't sure how to do this properly. I understand that `.body()` returns a response and I can't reuse the response builder?

pjenvey: Not jank at all, this is actix-web's way of doing this.

This should also set a json content-type:

```suggestion
                    builder
                        .content_type("application/json")
                        .body(EMPTY_TILES.as_str())
```

I thought `test::read_body_json` checked for such a content-type but apparently not. Let's add a check for it in the new test (a few of the other tests already have this code). We could consider adding a `read_body_json_strict` function that does both, but that's not critical for this PR.

pjenvey (changes requested):

pjenvey (approved):

PR #488, opened 2022-12-14
chore: tag 1.10.0

tiftran:


ncloudioj:
Let's also update the versions in `test-engineering` [here](https://github.com/mozilla-services/contile/blob/main/test-engineering/contract-tests/volumes/client/scenarios.yml#L501) and [here](https://github.com/mozilla-services/contile/blob/main/test-engineering/contract-tests/volumes/client/scenarios_tiles_cache.yml#L91) (there are a couple of those in `scenarios_tiles_cache.yml`).

ncloudioj (approved):
:shipit:

PR #496, opened 2023-01-09
fix: adm update metrics to be consistent

tiftran:
Currently, we only have metrics for updates for changes that occur for the adm filters. This made it hard to determine whether there was an attempt at an update or whether there were no changes. Added metrics for check attempts.

In addition, changed metric from adm.filter to filter.adm for consistency

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -474,13 +474,18 @@ impl AdmFilter {
 
 #[cfg(test)]
 mod tests {
+    use std::sync::Arc;
+    use std::time::Duration;
+    use actix_web::rt;
     use super::{check_url, AdmFilter};
     use crate::adm::settings::AdmAdvertiserSettings;
-    use crate::adm::AdmDefaults;
+    use crate::adm::{AdmDefaults, spawn_updater};
     use crate::adm::{settings::AdvertiserUrlFilter, tiles::AdmTile};
     use crate::tags::Tags;
     use crate::web::test::find_metrics;
-    use cadence::SpyMetricSink;
+    use cadence::{SpyMetricSink, StatsdClient};
+    use tokio::sync::RwLock;
+    use url::Url;
 
     #[test]
     fn check_url_matches() {
@@ -742,9 +747,7 @@ mod tests {
             .check_image_hosts(&defaults, &mut tile, &mut tags)
             .is_ok());
     }
-}
-
-#[actix_web::test]
+    #[actix_web::test]
 async fn check_advertiser_metrics() {
     let s = r#"{"adm_advertisers":{
             "Acme": {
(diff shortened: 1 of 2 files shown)


tiftran:
i feel like there should probably be tests for this? not sure how to write them though, couldn't find anything existing i could work off from either...

src/adm/filter.rs
```diff
@@ -121,12 +121,13 @@ pub fn spawn_updater(
         let mut tags = crate::tags::Tags::default();
         loop {
             {
+                metrics.incr("filter.adm.update_check").ok();
```
ncloudioj, line 124: I'd recommend injecting metrics into every arm of this match expression to get the whole picture for each update attempt.

```rs
    {
        match mfilter.read().await.requires_update(&storage_client).await {
            Ok(true) => {
                let mut filter = mfilter.write().await;
                match filter.update(&storage_client).await {
                    Ok(_) => {
                        metrics.incr("adm.filter.update.ok").ok();
                    }
                    Err(e) => {
                        filter.report(&e, &mut tags);
                        metrics.incr("adm.filter.update.error").ok();
                    }
                }
                // filter.update(&storage_client).await.unwrap_or_else(|e| {
                // metrics.incr("adm.filter.update.error").ok();
                // filter.report(&e, &mut tags);
                // });
            }
            Ok(false) => {
                metrics.incr("adm.filter.update.check.skip").ok();
            }
            Err(e) => {
                mfilter.read().await.report(&e, &mut tags);
                metrics.incr("adm.filter.update.check.error").ok();
            }
        }
    }
```
What do you think?

ncloudioj:
> i feel like there should probably be tests for this? not sure how to write them though, couldn't find anything existing i could work off from either...

Looks like we don't have any tests for metrics in the unit/integration tests. So your PR is fine :)

FWIW, Merino-rs uses a logging sink for metrics in tests, perhaps we can do the same for Contile in the future.

src/adm/filter.rs
tiftran: thorough sounds good

pjenvey:
Contile does have a couple uses of the `SpyMetricsSink` you could test this with, see integration tests (`init_app_with_spy` in `web/test.rs`) and the `img_storage` module also has another example.

ncloudioj (approved):
Looks good. Looks like CI busted. We might need to add `RUSTSEC-2021-0141` to https://github.com/mozilla-services/contile/blob/main/.cargo/audit.toml. Longer term, let's see if we can drop `dotenv`.

@pjenvey pointed out that `SpyMetricsSink` was available for testing, can you try leveraging it for some test coverage?

tiftran:
created a separate pr for ignoring dotenv rustsec https://github.com/mozilla-services/contile/pull/497

ncloudioj (approved):

tiftran:
eek had import troubles cause this dummy didn't place the test inside the test module. @pjenvey thank you for helping silly me

pjenvey (approved):

PR #504, opened 2023-01-14
fix: reduce the work done while holding the updater write lock

### LINKED ISSUE #502: Reduce the work done while holding the AdmFilter updater write lock
The background updater that periodically reads settings produced from shepherd currently holds the write lock, used when updating, much longer than it needs to – it’s actually making the network call to cloud storage while it’s acquired.

It should instead do the bare minimum while holding the write lock to avoid blocking the reads for too long.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2203)

### LINKED ISSUE #503: AdmFilter updater corrupts the all_include_regions Set
There’s a bug in the update code where it clears the all_include_regions (used as an optimization to return empty tiles to countries without any tiles) too often, resulting in it only containing one (instead of the 8 or so we expect there) country when the update’s completed.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/DISCO-2204)

pjenvey:
also fix a bug in the updater's handling of all_include_regions

DISCO-2203 & DISCO-2204

Closes #502
Closes #503

### DIFF
--- a/src/adm/filter.rs
+++ b/src/adm/filter.rs
@@ -112,39 +112,39 @@ pub fn spawn_updater(
             return Ok(());
         }
     }
-    let mfilter = filter.clone();
+    let mfilter = Arc::clone(filter);
     rt::spawn(async move {
-        let mut tags = crate::tags::Tags::default();
         loop {
-            {
-                // Do the check inside of a scope so that the read lock can be released right away.
-                let should_update = { mfilter.read().await.requires_update(&storage_client).await };
-                match should_update {
-                    Ok(true) => {
-                        let mut filter = mfilter.write().await;
-                        match filter.update(&storage_client).await {
-                            Ok(_) => {
-                                metrics.incr("filter.adm.update.ok").ok();
-                            }
-                            Err(e) => {
-                                filter.report(&e, &mut tags);
-                                metrics.incr("filter.adm.update.error").ok();
-                            }
-                        }
-                    }
-                    Ok(false) => {
-                        metrics.incr("filter.adm.update.check.skip").ok();
-                    }
-                    Err(e) => {
-                        mfilter.read().await.report(&e, &mut tags);
-                        metrics.incr("filter.adm.update.check.error").ok();
-                    }
-                }
-            }
+            updater(&mfilter, &storage_client, &metrics).await;
             rt::time::sleep(refresh_rate).await;
         }
     });
     Ok(())
+}
+
+/// Update `AdmFilter` from the Cloud Storage settings if they've been updated
+async fn updater(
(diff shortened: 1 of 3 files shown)


src/adm/filter.rs
```diff
@@ -112,41 +112,41 @@ pub fn spawn_updater(
...
 
+/// Update `AdmFilter` from the Cloud Storage settings if they've been updated
+async fn updater(
+    filter: &Arc<RwLock<AdmFilter>>,
+    storage_client: &cloud_storage::Client,
+    metrics: &Arc<StatsdClient>,
+) {
+    // Do the check before matching so that the read lock can be released right away.
+    let result = filter.read().await.fetch_new_settings(storage_client).await;
+    match result {
```
jrconlin, line 133: we could probably combine these two.

jrconlin (approved):

src/adm/filter.rs
```diff
@@ -112,41 +112,41 @@ pub fn spawn_updater(
...
 }
 
+/// Update `AdmFilter` from the Cloud Storage settings if they've been updated
+async fn updater(
+    filter: &Arc<RwLock<AdmFilter>>,
+    storage_client: &cloud_storage::Client,
+    metrics: &Arc<StatsdClient>,
+) {
+    // Do the check before matching so that the read lock can be released right away.
+    let result = filter.read().await.fetch_new_settings(storage_client).await;
```
ncloudioj, line 132: Ah, I didn't know that calling `fetch_new_settings(&self)` would also drop the lock guard from the current scope. TIL!

src/adm/filter.rs
```diff
@@ -182,44 +183,40 @@ impl AdmFilter {
...
+            for country in setting.keys() {
+                self.all_include_regions.insert(country.clone());
             }
-            self.last_updated = Some(chrono::Utc::now());
+            self.advertiser_filters
+                .adm_advertisers
+                .insert(adv.to_lowercase(), setting);
         }
-        Ok(())
+        self.last_updated = Some(chrono::Utc::now());
```
ncloudioj, line 219: Instead of the current timestamp, we might want to store the last modified timestamp of the remote file to avoid some subtle ambiguities because `filter::update()` might be done significantly after `filter::fetch_new_settings()` (e.g. due to a long wait for the write lock). If there is any new update published between `filter:fetch_new_settings()` and `filter:update()`, Contile will be unable to pick it up at the next update check due to that `obj.updated < self.last_updated` check in `fetch_new_settings()`.

We're not making that much publish at the moment, so this can be handled later with a follow-up.

ncloudioj (approved):
LGTM

PR #505, opened 2023-01-16
fix: use last object modification timestamp for filter updater

ncloudioj:
The follow-up of #504 to mitigate [this anomaly](https://github.com/mozilla-services/contile/pull/504#discussion_r1071539794).


pjenvey (approved):
👍 I was thinking the same thing and almost did this in the other PR 😄

PR #519, opened 2023-02-24
[DISCO-2250] feat: Return 204 for unknown form factors

ncloudioj:
Per the partner's request, we don't want to serve tiles for clients if their form factor (device type) is unknown. 

This fixes [DISCO-2250](https://mozilla-hub.atlassian.net/browse/DISCO-2250).

[DISCO-2250]: https://mozilla-hub.atlassian.net/browse/DISCO-2250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/web/handlers.rs
+++ b/src/web/handlers.rs
@@ -36,8 +36,7 @@ pub async fn get_tiles(
     trace!("get_tiles");
     metrics.incr("tiles.get");
 
-    let settings = &state.settings;
-    if let Some(response) = maybe_early_respond(&settings, &state, &location, &device_info).await {
+    if let Some(response) = maybe_early_respond(&state, &location, &device_info).await {
         return Ok(response);
     }
     let audience_key = cache::AudienceKey {
@@ -53,6 +52,7 @@ pub async fn get_tiles(
         legacy_only: device_info.legacy_only(),
     };
 
+    let settings = &state.settings;
     let mut tags = Tags::from_head(request.head(), settings);
     {
         tags.add_extra("audience_key", &format!("{:#?}", audience_key));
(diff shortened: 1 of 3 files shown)


src/web/handlers.rs
```diff
@@ -37,24 +37,7 @@ pub async fn get_tiles(
...
-        // optionally send 200 to resolve
-        // https://github.com/mozilla-services/contile/issues/284
-        let response = if settings.excluded_countries_200 {
-            HttpResponse::Ok()
-                .content_type("application/json")
-                .body(EMPTY_TILES.as_str())
-        } else {
-            HttpResponse::NoContent().finish()
-        };
+    if let Some(response) = maybe_early_respond(&settings, &state, &location, &device_info).await {
```
linabutler, line 40: praise: Nice refactoring! 🙏🏼

src/web/handlers.rs
```diff
@@ -212,3 +195,46 @@ fn fallback_response(settings: &Settings, tiles: &cache::Tiles) -> HttpResponse
...
+            HttpResponse::Ok()
+                .content_type("application/json")
+                .body(EMPTY_TILES.as_str())
+        } else {
+            HttpResponse::NoContent().finish()
+        };
+        return Some(response);
+    }
+
+    if matches!(&device_info.form_factor, FormFactor::Other) {
```
linabutler, line 234: suggestion (non-blocking): Let's do this check first, before the live markets check above? That one looks more involved, and this is a simpler pattern match.

src/web/handlers.rs
```diff
@@ -212,3 +195,46 @@ fn fallback_response(settings: &Settings, tiles: &cache::Tiles) -> HttpResponse
...
+/// Check if the tile request should be responded early.
+///
+/// This allows us to short circuit requests if:
+///   - they are not sent from regions of the live markets
+///   - they are sent from unknown device types (`form_factor == "other"`)
+///
+/// It returns a proper response if the early response is desired.
+/// Otherwise, it returns None.
+async fn maybe_early_respond(
+    settings: &Settings,
```
linabutler, line 208: suggestion (non-blocking): Can you use `state.settings` to get at the settings, instead of passing it as a separate argument?

linabutler (approved):
Looks great—thanks! 🥳

linabutler (approved):
Perfect!

PR #522, opened 2023-03-02
fix: Fix intermittent contract test failures on M1 laptops

ncloudioj:
Sigh, turns out the test failures were caused by the crappy network latency issues (only in docker-compose) on M1 laptops. Increasing both the connection and adm_request timeouts seemed to work, though it will take a little bit longer to finish all the tests, which is still better than failures.

cc: @Trinaa


Trinaa (approved):
I can't thank you enough for finding a fix to this @ncloudioj!

PR #530, opened 2023-04-11
Add some commands to run Coverage and Checks locally

mtbottle:
This change does 2 things

1. Adding a Makefile to run checks and create a coverage report
2. Add some tests to a random file (cache)

This is a first step to getting us higher test coverage to prepare Contile for Rapid Release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,13 @@
-build-dev:
-	cargo build --tests
+build:
+	cargo build
 
-setup-grcov:
-	cargo install grcov
+setup-all: build setup-coverage-tools setup-rust-checks
 
-coverage: build-dev setup-grcov
-	RUSTFLAGS="-C instrument-coverage" cargo test
-	grcov . -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existing -o ./target/debug/coverage/
-	rm *.profraw
+setup-coverage-tools:
+	cargo +stable install cargo-llvm-cov --locked
+
+coverage: setup-coverage-tools
+	cargo llvm-cov --open
 
 setup-rust-checks:
 	rustup component add rustfmt
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+build-dev:
+	cargo build --tests
+
+setup-grcov:
+	cargo install grcov
+
+coverage: build-dev setup-grcov
+	RUSTFLAGS="-C instrument-coverage" cargo test
+	grcov . -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existing -o ./target/debug/coverage/
+	rm *.profraw
+
+setup-rust-checks:
+	rustup component add rustfmt
+	cargo install cargo-audit
+	rustup component add clippy
+
+checks: setup-rust-checks
+	cargo fmt -- --check
+	cargo audit
+	cargo clippy --all --all-targets --all-features -- -D warnings
--- a/src/server/cache.rs
+++ b/src/server/cache.rs
@@ -309,19 +309,40 @@ async fn tiles_cache_periodic_reporter(cache: &TilesCache, metrics: &Metrics) {
 
 #[cfg(test)]
 mod test_tile_cache {
-
     use super::TilesCache;
     use crate::server::TILES_CACHE_INITIAL_CAPACITY;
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,20 @@
+build-dev:
+	cargo build --tests
```
ncloudioj, line 2: Looks like `build-dev` is only used by `coverage`, you can just do `RUSTFLAGS="-Cinstrument-coverage" cargo test --all-targets`. `--all-targets` includes `--tests`, also it will automatically build all the libs and bins before running all the tests.

Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+setup-grcov:
+	cargo install grcov
```
ncloudioj, line 5: Do we also need to install `rustup component add llvm-tools-preview`?

Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+coverage: build-dev setup-grcov
+	RUSTFLAGS="-C instrument-coverage" cargo test
+	grcov . -s . --binary-path ./target/debug/ -t html --branch --ignore-not-existing -o ./target/debug/coverage/
+	rm *.profraw
```
ncloudioj, line 10: Maybe add `*.profraw` to gitignore and dockerignore as well?

Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+setup-rust-checks:
+	rustup component add rustfmt
+	cargo install cargo-audit
+	rustup component add clippy
```
ncloudioj, line 15: Shall we add a top-level `setup` to include all the sub-setup targets here?

src/server/cache.rs
```diff
@@ -306,3 +306,67 @@ async fn tiles_cache_periodic_reporter(cache: &TilesCache, metrics: &Metrics) {
     metrics.count("tiles_cache.count", cache_count);
     metrics.count("tiles_cache.size", cache_size as i64);
 }
+
+#[cfg(test)]
+mod test_tile_cache {
+
```
ncloudioj, line 312:
```suggestion
```

src/server/cache.rs
```diff
@@ -306,3 +306,67 @@ async fn tiles_cache_periodic_reporter(cache: &TilesCache, metrics: &Metrics) {
...
+    use cadence::{NopMetricSink, StatsdClient};
+    use std::{sync::Arc, time::Duration};
+
+    #[actix_web::test]
+    async fn test_spawn_periodic_reporter() {
+        let tiles_cache = TilesCache::new(TILES_CACHE_INITIAL_CAPACITY);
+        let statsd_client = StatsdClient::builder("test", NopMetricSink).build();
+
+        // Just checks that we can spawn the periodic reporter without issues.
+        tiles_cache.spawn_periodic_reporter(Duration::from_secs(1), Arc::new(statsd_client));
```
ncloudioj, line 324: There are no assertions here, maybe we can check those metric counters (`tiles_cache.count` and `tiles_cache.size`) emitted inside the reporter. See [here](https://github.com/mozilla-services/contile/blob/main/src/web/test.rs#L902-L941) for an example of how we use `SpyMetricSink` for that.

Makefile
mtbottle: Ah whoops! Thanks for the catch!

pjenvey:
I'll note there's also the [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov) crate now which appears to be a maybe more convenient alternative to grcov. It's been suggested it [works better](https://github.com/mozilla/grcov/pull/533#issuecomment-1106594405)/[is more accurate](https://github.com/taiki-e/cargo-llvm-cov/issues/134#issuecomment-1032162970) but I'm wondering if that's due to it forcing `RUST_TEST_THREADS=1` to avoid a [known issue](https://github.com/taiki-e/cargo-llvm-cov#known-limitations) w/ `-Cinstrument-coverage` If we're seeing intermittent differences in coverage w/ grcov also we'll probably need to force it as well.

mtbottle:
I've switched to using `cargo-llvm-cov` because I think it's simpler and cleaner to use.

ncloudioj (approved):
LGTM, thanks!

PR #532, opened 2023-04-18
Chore/update 0323

jrconlin:


.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ commands:
       - run:
           name: Rust Clippy
           command: |
-            cargo clippy --all --all-targets --all-features -- -D warnings
+            cargo clippy --all --all-targets --all-features -- -D warnings -A clippy::result_large_err
```
ncloudioj, line 42: Could you also add this to the [one](https://github.com/mozilla-services/contile/blob/main/Makefile#L20) in the Makefile?

ncloudioj (dismissed):
Newer Rust, YAY!

Thanks JR!

.circleci/config.yml
ncloudioj: Looks like we need to wrap `tags` into a Box as well. I've pushed a fix on top of JR's patch. Mind taking a quick look?

ncloudioj:
Thank you both! Now merging those goodies 🚢

PR #533, opened 2023-04-24
feat: Add a metric to measure the tile count of adm response

ncloudioj:
This closes [DISCO-2346](https://mozilla-hub.atlassian.net/browse/DISCO-2346).

r? @pjenvey for extra eyeballs since this is the first time to use a distribution metric in Contile.

[DISCO-2346]: https://mozilla-hub.atlassian.net/browse/DISCO-2346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/adm/tiles.rs
+++ b/src/adm/tiles.rs
@@ -245,6 +245,12 @@ pub async fn get_tiles(
                 })?
         }
     };
+    metrics.count_with_tags(
+        "tiles.adm.response.tiles_count",
+        response.tiles.len() as i64,
+        Some(tags),
+    );
+    // Keep this counter to facilitate the "zero" ad fill rate monitoring
     if response.tiles.is_empty() {
         warn!("adm::get_tiles empty response {}", adm_url);
         metrics.incr_with_tags("filter.adm.empty_response", Some(tags));
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -118,9 +118,7 @@ impl Metrics {
     /// Duration is calculated when this timer is dropped.
     pub fn start_timer(&mut self, label: &str, tags: Option<Tags>) {
         let mut mtags = self.tags.clone().unwrap_or_default();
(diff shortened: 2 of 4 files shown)


ncloudioj:
We could either increment a counter such as `contile.tiles.adm.response` with a new tag `tile.count`, but that will add this tag to all the subsequent metrics which would be wasteful, or use a plain counter `contile.tiles.adm.response.tile_count`, but that'll be not super easy to use for monitoring & analysis.

src/adm/tiles.rs
```diff
@@ -245,6 +245,12 @@ pub async fn get_tiles(
                 })?
         }
     };
+    metrics.distribution_with_tags(
+        "tiles.adm.response.distribution",
```
pjenvey, line 249: I don't care if 'distribution' is included in this name but maybe something more specific:
```suggestion
        "tiles.adm.response.tiles_count",
```

pjenvey (commented):
@ncloudioj FYI the distribution type isn't supported by telegraf's default configuration, however it can parse them when both these are set: `datadog_extensions = true` (which we already have) and `datadog_distributions = true` (which we don't).

So this will require to SRE add that.

ncloudioj:
Thanks, Phil! I will double check with SRE to see if those changes are viable. If not, will reformat this to other metric types.

ncloudioj:
Per SRE, we're using a pretty old version of InfluxDB so its support of the distribution metric type is unclear. Also, this work is due on May 5th, so let's use a counter metric for this for now.

pjenvey (approved):
Count obviously isn't ideal here obviously but I guess we can at least look at the breakdown of counts over small intervals tagged by e.g. geo country code.

PR #529, opened 2023-04-06
feat: Add an `adm_mobile_max_tiles` setting (DISCO-2306)

linabutler:
This commit adds an option to separately control the maximum number of tiles returned to Desktop and mobile clients.

Desktop clients use the `adm_max_tiles` setting now, and mobile clients use `adm_mobile_max_tiles` (falling back to `adm_max_tiles`).

PR #539, opened 2023-05-02
Revert "feat: Add an `adm_mobile_max_tiles` setting (DISCO-2306, #529)"

ncloudioj:
This reverts commit 23f5ae6993ed93744bf92bfb3bbca3c900a1eaeb.

We will wait for the dismissal telemetry to hit the road in Fx 113 to enable this change in Contile.

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -103,14 +103,12 @@ pub struct Settings {
     pub adm_sub1: Option<String>,
     /// adm Endpoint URL
     pub adm_endpoint_url: String,
-    /// max number of tiles returned to non-mobile clients (default: 3)
-    pub adm_max_tiles: u8,
     /// Mobile versions of the above
     pub adm_mobile_partner_id: Option<String>,
     pub adm_mobile_sub1: Option<String>,
     pub adm_mobile_endpoint_url: Option<String>,
-    /// max number of tiles returned to mobile clients (default: 2)
-    pub adm_mobile_max_tiles: Option<u8>,
+    /// max number of tiles returned to clients (default: 2)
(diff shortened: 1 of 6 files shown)


pjenvey (commented):
Alternatively we could set both settings to the original 3 value in the helm chart.

src/settings.rs
```diff
@@ -103,14 +103,12 @@ pub struct Settings {
     pub adm_sub1: Option<String>,
     /// adm Endpoint URL
     pub adm_endpoint_url: String,
-    /// max number of tiles returned to non-mobile clients (default: 3)
-    pub adm_max_tiles: u8,
```
linabutler, line 107: I am a little surprised that it's easier to make a code change than to set an environment variable to override this, though. Isn't this setting controlled by `CONTILE_ADM_MAX_TILES`? How come we can't set it to `2` to override the new default of 3?

ncloudioj:
You know what, it's very likely that we will bump from 3 to 4 or other values in the future, we will be using the EnvVar for that anyway. I will file a PR to add that to the cloud-infra repo instead.

Thanks for the reminder.

pjenvey:
@ncloudioj FYI we (the contile-reviewers group) do now have ownership (can approve) files under the `topsites/` directory so they don't even need approval. I think this settings addition hasn't rolled out to stage/prod yet so I think we just need to remind them to incorporate the helm change for the next release (which might even happen automatically once it's merged).

ncloudioj:
kk, close this to favor https://github.com/mozilla-services/cloudops-infra/pull/4853.

PR #551, opened 2023-05-09
[DISCO-2383] feat: Use SHA commit hash as version

ncloudioj:
## References

JIRA: [DISCO-2383](https://mozilla-hub.atlassian.net/browse/DISCO-2383)
GitHub: N/A

## Description
This patch replaces all the version strings from Cargo.toml in Contile with `$CIRCLECI_SHA1` including Sentry, logging, Dockerflow, and outbound traffic. Note that it uses the short SHA1 (the leading 7 characters) rather than the original 40 characters string. 

## PR Review Checklist

- [x] Functional and performance test coverage has been expanded and maintained 

[DISCO-2383]: https://mozilla-hub.atlassian.net/browse/DISCO-2383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -148,7 +148,10 @@ jobs:
       - write-version
       - run:
           name: Build Docker image
-          command: docker build -t app:build .
+          command: |
+            docker build \
+              -t app:build \
+              --build-arg VERSION="$(echo ${CIRCLE_SHA1} | cut -c -7)" .
```
ncloudioj, line 154: Note that I opted to use the short version of SHA1 here as it's identifying enough and can save a lot of space.

Dockerfile
```diff
@@ -19,7 +23,7 @@ RUN \
     cargo --version && \
     rustc --version && \
     mkdir -m 755 bin && \
-    cargo build --release && \
+    CONTILE_VERSION=${VERSION} cargo build --release && \
```
ncloudioj, line 26: FWIW, `$CONTILE_VERSION` is only used at compile time, so it's only referenced when we build Contile.

src/web/dockerflow.rs
```diff
@@ -52,10 +52,7 @@ async fn version() -> HttpResponse {
 /// Returns a status message indicating the current state of the server
 async fn heartbeat(state: web::Data<ServerState>) -> HttpResponse {
     let mut checklist = HashMap::new();
-    checklist.insert(
-        "version".to_owned(),
-        Value::String(env!("CARGO_PKG_VERSION").to_owned()),
-    );
+    checklist.insert("version".to_owned(), Value::String(create_app_version("/")));
```
ncloudioj, line 55: Note that apart from `CARGO_PKG_VERSION`, `CARGO_PKG_NAME` is also included in this field which I assume should be fine to do here.

src/main.rs
```diff
@@ -1,5 +1,6 @@
 //! Main application entry point
 #![forbid(unsafe_code)]
+use std::borrow::Cow;
```
Trinaa, line 3: 🐄 ?

Trinaa:
In the CircleCI config `write-version` command we assign 'CIRCLE_TAG' to the 'version' field, should that be empty now?

src/main.rs
ncloudioj: Lol, it's a shorthand for `Clone on Write`.

ncloudioj:
> In the CircleCI config write-version command we assign 'CIRCLE_TAG' to the 'version' field, should that be empty now?

Per Circle's doc, it will be an empty string if the build was not tagged (which we will be doing onwards), so we're good here. FWIW, Merino uses the same for `write-version` and it returns.

```sh
$ curl https://merino.services.mozilla.com/__version__ | jq .

{
  "source": "https://github.com/mozilla-services/merino-py",
  "version": "",
  "commit": "4dd482f7faf1a6667c4770da85850bd463b64232",
  "build": "https://circleci.com/gh/mozilla-services/merino-py/8433"
}
```

src/main.rs
Trinaa: Love it!

Trinaa (approved):

PR #561, opened 2023-05-13
[DISCO-2394] chore: Use cargo chef to speed up docker build

ncloudioj:
## References

JIRA: [DISCO-2394](https://mozilla-hub.atlassian.net/browse/DISCO-2394)
GitHub: N/A

## Description
Use Cargo Chef to speed up docker build. Empirical results suggest a 50% reduction in docker build duration. 

## PR Review Checklist

- [ ] [Functional and performance test](https://github.com/mozilla-services/contile/tree/main/test-engineering) coverage has been expanded and maintained (if applicable)
- [ ] `[do not deploy]` and `[load test: (abort|warn)]` keywords are applied (if applicable)


pjenvey:
Seems like it's very hit or miss but when it works it works well.

It's odd that the last couple of builds it did cache hit the initial docker steps, but for whatever reason not the cargo chef cook step where it really counts.

Wanna just merge it and we'll monitor it over time (probably want to add a [do not deploy])?

ncloudioj:
Thanks, Phil! Indeed, it works really well for local builds, and vastly cuts the build time so long as the dependencies remain unchanged, so I think that's already a win. However it's quite hit or miss in CI, I guess it has to do with how CircleCI's container caching layer works. I've tried to overcome that but still couldn't get it to leverage the dependency cache all the time. The bottom line is that it doesn't regress the build time in CI 🤷‍♂️.

> Wanna just merge it and we'll monitor it over time (probably want to add a [do not deploy])?
Yup, let's give it a spin and see how useful it can be in practice. re: [do not deploy], we haven't fully set up the CD bit in Jenkins, so it will just publish docker images without deploying :)

pjenvey (approved):

PR #611, opened 2023-07-13
[DISCO-2516] fix: Return HTTP 204 for excluded regions

ncloudioj:
## References

JIRA: [DISCO-2516](https://mozilla-hub.atlassian.net/browse/DISCO-2516)
GitHub: N/A

## Description

Firefox now can handle HTTP 204 properly, Contile can return HTTP 204 for excluded regions. This flips the pref added in #290.

## PR Review Checklist

- [x] [Functional and performance test](https://github.com/mozilla-services/contile/tree/main/test-engineering) coverage has been expanded and maintained (if applicable)
- [ ] The `[load test: (abort|warn)]` keywords are applied (if applicable)
- [ ] The PR title starts with the JIRA issue reference, format `[DISCO-####]`, and has the same title

[DISCO-2516]: https://mozilla-hub.atlassian.net/browse/DISCO-2516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -169,7 +169,7 @@ impl Default for Settings {
             exclude_dma: Some("[798, 583, 740]".to_owned()),
             connect_timeout: 2,
             request_timeout: 5,
-            excluded_countries_200: true,
+            excluded_countries_200: false,
             cache_control_header: true,
             // ADM specific settings
             adm_endpoint_url: "".to_owned(),
--- a/src/web/test.rs
+++ b/src/web/test.rs
@@ -738,6 +738,7 @@ async fn empty_tiles_returns_no_content() {
     let adm_settings = filters;
     let mut settings = Settings {
         adm_endpoint_url: adm.endpoint_url,
+        excluded_countries_200: true,
         adm_settings,
         ..get_test_settings()
     };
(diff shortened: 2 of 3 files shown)


src/settings.rs
```diff
@@ -169,7 +169,7 @@ impl Default for Settings {
             exclude_dma: Some("[798, 583, 740]".to_owned()),
             connect_timeout: 2,
             request_timeout: 5,
-            excluded_countries_200: true,
+            excluded_countries_200: false,
```
linabutler, line 172: It sounds like this setting was introduced to work around a bug in older versions of Firefox. Shall we get a follow-up on file to remove the setting entirely? 😄

linabutler (approved):
Stamp!

src/settings.rs
ncloudioj: Good point, let me file a ticket for that. We will deploy this change to prod and see how Firefox behaves. If nothing goes south, we can drop this treatment in contile.

ncloudioj: Here is the ticket: https://mozilla-hub.atlassian.net/browse/DISCO-2518.

linabutler (approved):

Issue #613, opened 2023-07-18
Contile: Add local development set-up documentation


taddes:
Currently this documentation exists in the runbook. However, it really belongs in in the GitHub repo.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/DISCO-2536)

PR #614, opened 2023-07-18
[DISCO-2488] Contile: Add local development set-up documentation

taddes:
## References

JIRA: [DISCO-2488](https://mozilla-hub.atlassian.net/browse/DISCO-2488)
GitHub: [#613 ](https://github.com/mozilla-services/contile/issues/613)

## Description
Currently this documentation exists in the runbook. However, it really belongs in in the GitHub repo.

Acceptance Criteria:
Add the above instructions to the `README.md` file in the root of the repo. 

## PR Review Checklist

- [ ] This PR conforms to the [Opsec policies](https://github.com/mozilla-services/websec-check)
- [ ] [Functional and performance test](https://github.com/mozilla-services/contile/tree/main/test-engineering) coverage has been expanded and maintained (if applicable)
- [ ] The `[load test: (abort|warn)]` keywords are applied (if applicable)

[DISCO-2488]: https://mozilla-hub.atlassian.net/browse/DISCO-2488?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


README.md
```diff
@@ -23,14 +23,73 @@ practices.
...
+
+adm_endpoint_url="http://localhost:5000/tilesp/desktop"
+adm_defaults='{"image_hosts":["example.com","example.org"],"impression_hosts":["example.com","example.org"],"click_hosts":["example.com","example.org"]}'
+adm_settings="test-engineering/contract-tests/volumes/contile/adm_settings.json"
+adm_partner_id="partner_id_test"
+adm_sub1="sub1_test"
+adm_has_legacy_image='["Example ORG","Example COM"]'
+adm_timeout=2
+tiles_ttl=0
+```
```
tiftran, line 77: I believe the contile gets cranky and doesn't start up without a sov source. You would need to add `sov_source`  probably easiest to use the the contract test example: https://github.com/mozilla-services/contile/blob/main/test-engineering/contract-tests/volumes/contile/sov_settings.json

taddes: Great, thank you for mentioning that @tiftran , I'll make sure to update here and in the runbook as it's not specified there either.

README.md
```diff
@@ -23,14 +23,75 @@ practices.
...
+    -v `pwd`/test-engineering/contract-tests/volumes/partner:/tmp/partner \
+    -p 5000:5000 \
+    mozilla/contile-integration-tests-partner
+```
+
+To use the above, you'll need to set the Contile configuration option as follows:
+`adm_endpoint_url`  = `http://localhost:5000/tilesp/desktop`.  Note, you also need to 
+define a SOV (Share of Voice) source file, so make sure this is provided for partner allocations. Currently, a sample file is in the contract-tests directory, so you can make use of that, or define your own allocation JSON file to point to. 
+
+It may be helpful to create a settings file to contain all of the preferred settings, like the example of `sa-test.toml` below:
```
mtbottle, line 59: Just a thought on explanations: I would suggest that we reduce the amount of explanatory text for set up documentation. There's a few reasons for this:

1. Audience: users who use set up documentation usually have the goal to set up their service and care less about _how_ the pieces fit together. So a lot of this explanatory text will get in the way of their goal.
2. Explanatory text here will introduce another surface area for documentation maintenance (i.e. if we change Share of Voice related setup, then we'll need to change the explanation here too)
3. Giving users _options_ for set up will actually be more confusing and make it longer for users to actually get something set up properly.

The [Application Service Build set up](https://mozilla.github.io/application-services/book/building.html#macos) is a good example of how we should write set up documentation. Some things to note:

1. Code commands can be copy and pasted directly into the terminal
2. Instructions are short and straight to the point (don't explain the actual set up or the reason for requiring it, as it's not important to getting it actually set up).
3. Sequencing for steps are clear and numbered. If users have trouble, it's much easier to ask for help with the reference "Step X" rather than "the point I install XYZ".

README.md
```diff
@@ -23,14 +23,75 @@ practices.
...
+
+```toml
```
mtbottle, line 61: Since this "code" is going to be put into a file anyways, what do you think about just adding a `sa-test.toml` file to the repo for easier set up?

mtbottle (changes requested):
A few comments.

README.md
taddes: Sounds good @mtbottle , I agree.  Shall I update the runbook to match the same tone and reduce verbiage so the two match? I was going off of it for this update, so it's probably best if the match, right?

mtbottle: Ah yes! The original runbook should also be updated to match! Thanks for fixing that up as well!

README.md
```diff
@@ -23,13 +23,37 @@ practices.
...
+    ```shell
+        docker run \
+        --env PORT=5000 \
+        --env RESPONSES_DIR=/tmp/partner/ \
+        --env ACCEPTED_MOBILE_FORM_FACTORS=phone,tablet \
+        --env ACCEPTED_DESKTOP_FORM_FACTORS=desktop \
+        -v `pwd`/test-engineering/contract-tests/volumes/partner:/tmp/partner \
+        -p 5000:5000 \
+        mozilla/contile-integration-tests-partner
+    ```
```
mtbottle, line 43: nit to make this more concise:
```suggestion
3. Start connection a local ADM instance:
    ```shell
        docker run \
        --env PORT=5000 \
        --env RESPONSES_DIR=/tmp/partner/ \
        --env ACCEPTED_MOBILE_FORM_FACTORS=phone,tablet \
        --env ACCEPTED_DESKTOP_FORM_FACTORS=desktop \
        -v `pwd`/test-engineering/contract-tests/volumes/partner:/tmp/partner \
        -p 5000:5000 \
        mozilla/contile-integration-tests-partner
    ```
``

File: README.md [hunk 5]
```diff
@@ -23,13 +23,37 @@ practices.
...
+4. Configuration settings are contained in the sample `sa-test.toml` file at the root of the Contile repository.  
+You may change settings [there](sa-test.toml) that pertain to your local development on Contile.
+5. Start the local dev application by running: 
+```shell
+ #! /bin/bash
+RUST_LOG=contile=trace,config=debug \
+    cargo run -- --config sa-test.toml #--debug-settings
+```
```
mtbottle, line 51: Step 4 is kind of unnecessary. Someone reading the command can probably extrapolate that `sa-test.toml` contains test configurations.
```suggestion
4. Start the local dev application by running: 
```shell
 #! /bin/bash
RUST_LOG=contile=trace,config=debug \
    cargo run -- --config sa-test.toml #--debug-settings
```
```

File: README.md [hunk 6]
```diff
@@ -23,13 +23,37 @@ practices.
...
+6. If you used the sample `sa-test.toml` configuration file for settings, you should
+be able to make requests to the dev server like so:
```
mtbottle, line 53: Command below is good! Just a suggestion to make this less verbose and more maintainable:

```suggestion
6. Check that the service can accept requests by running:
```

README.md
```diff
@@ -23,13 +23,37 @@ practices.
...
-## Setting Up
-
-Contile uses Rust, and requires the latest stable iteration. See
-[rustup.rs](https://rustup.rs/) for how to install this application.
-
-Once Rust is installed you can compile using `cargo build`. This will
-create a development release.
+## Setting Up Your Local Dev Environment
+1. Install Rust. See [rustup.rs](https://rustup.rs/) for how to install on your platform.
+2. Compile Contile using `cargo build`. This will create a development release.
```
mtbottle, line 28: nit: remove more unnecessary info. The title of this section is: "Setting Up Your Local Dev Environment" and "development release" is ambiguous. 
```suggestion
2. Compile Contile using `cargo build`. 
```

mtbottle (changes requested):
Almost there! Just a few more suggestions to make the instructions more concise.

README.md
taddes: If it's ok with you, I'd kind of want to this one to stay. It's just a small flag in the `cargo run` command, which someone might not immediately see and I'd personally appreciate this context as a dev if I were getting set up. You ok if it remains? It read really nicely in the runbook

taddes:
Thanks @mtbottle , made requested changes, just suggested we leave one thing in about the toml file as I feel it adds some welcome clarity. I can remove it in the end if you prefer, I just tried to look at it as if I had 0 knowledge of Contile and I feel that adds some clarity with a single sentence.

README.md
mtbottle: Hmm, the extra step makes it look like you need to take some sort of action on step 4. The action that the user needs to take is unclear (i.e. is there something that they update something in the file? how do they know they've accomplished the step properly?). So, if we want to keep the extra context, let's merge the steps like:

```
  4. Start the local dev application with the configurations specified in `sa-test.toml`.
  ```shell
   #! /bin/bash
  RUST_LOG=contile=trace,config=debug \
      cargo run -- --config sa-test.toml 
  ``
  You may change settings that pertain to your local development on Contile directly in the file.
```

mtbottle (dismissed):
Generally LGTM. Just that comment about step 4. I think we should merge it with step 5 if you want to keep the context. However, I'll leave that decision to you, as I don't think it's a huge blocker.

taddes:
Hi @mtbottle , yeah! That sounds like a great middle ground, I'll merge it with step 5 so it adds a little context. Thanks

mtbottle (approved):
🚀

